PR #1, opened 2017-07-14
Allow suppression of pidfiles.

dkg:
When unbound is maintained by a system manager that can actually keep
track of what daemons are running correctly, a pidfile is an
opportunity for extra failures, rather than a way to avoid failures.

In particular, if unbound terminates roughly, it might leave the
pidfile lying around. Then a subsequent unbound instance will think
that another unbound process is running when it isn't (particularly if
some other stable process happens to have landed on the same process
ID).

So for a system manager that is already capable of keeping track of
whether the daemon is running or not, it would be nice to have the
ability to avoid these extra failure modes by just ignoring pidfiles
entirely.

Since the same unbound config file might need to be used on a
distribution that might use different system managers, being able to
control the use of a pidfile by a command-line argument is probably
the cleanest way to go (see https://bugs.debian.org/867192).

### DIFF
--- a/daemon/unbound.c
+++ b/daemon/unbound.c
@@ -400,7 +400,7 @@ detach(void)
 /** daemonize, drop user priviliges and chroot if needed */
 static void
 perform_setup(struct daemon* daemon, struct config_file* cfg, int debug_mode,
-	const char** cfgfile)
+	const char** cfgfile, int need_pidfile)
 {
 #ifdef HAVE_KILL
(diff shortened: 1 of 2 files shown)


ralphdolmans:
Closing this PR, was also submitted as patch to our bugzilla (https://www.nlnetlabs.nl/bugs-script/show_bug.cgi?id=1349) and is applied.

PR #2, opened 2018-03-08
support for redis cachedb backend

jinmeiib:
This change adds to unbound support for a new cached backend that uses a Redis server as the storage.

This implementation depends on the hiredis client library (https://redislabs.com/lp/hiredis/).  And unbound should be built with both --enable-cachedb and --with-libhiredis[=PATH] (where $PATH/include/hiredis/hiredis.h should exist).

I've locally confirmed the following:
- build succeeds with or without --with-libhiredis on Linux (Ubuntu 16.04.5), MacOS X 10.11.6 and FreeBSD 10.3.
- 'make test' passed (on Linux)
- 'make longtest' passed (on Linux)


chantra:
Thanks @jinmeiib ! that's great. I am really happy to see this getting in.
Would it make sense to try to keep the redis code in separate files (as much as possible) for the sake of separating implementation in their own files?

jinmeiib:
Thanks for the comment.  Yes, I think that makes sense especially if we expect more backend will be added.  I don't remember why I chose the single-file approach at that point, but perhaps it's because the size of redis-specific code isn't expected to be too big and I was not sure if we really have more backends and the speparation would be premature generalization.  Anyway, I'm now asking NLNet Labs people for a review.  If they want me to separate it, I'm happy to do that.  If they want to make it by themselves and merge it, that's also fine.  If they want to keep the current style for now, that's also fine for me.

wcawijngaards:
Hi Jinmei,

Thanks for the patch, I have added it!  Attributed to Jinmei Tatuya
(Infoblox).

Also separated out the code into cachedb/redis.c, as asked by chantra.

Best regards, Wouter

PR #3, opened 2018-03-15
fix cachedb portability problem

jinmeiib:
by moving htobe64/be64toh portability code from redis.c to cachedb.c.
the former actually doesn't use these functions and doesn't need them.
without the change cachedb doesn't compile on BSDs and MacOS X (and maybe some others).
(note that this is not related to the redis support.  the issue has been in cachedb.c from the beginning).


wcawijngaards:
Hi,

Applied this to fix that.  Thanks for the review of the code file split.

Best regards, Wouter

PR #5, opened 2018-04-30
Python module: define constant MODULE_RESTART_NEXT 

episource:
The python module does not define a constant `MODULE_RESTART_NEXT` for the `module_ext_state`'s enumeration value [`module_restart_next`](http://unbound.net/documentation/doxygen/module_8h.html#a42696d302cc419299b98207675a2ba18).

This pull requests adds the constant to swig's interface file.

I've been successfully using this patch with unbound-[1.6.0](https://github.com/episource/unbound/tree/dev/module_ext_state/v1.6.0), [1.6.1](https://github.com/episource/unbound/tree/dev/module_ext_state/v1.6.1), [1.6.4](https://github.com/episource/unbound/tree/dev/module_ext_state/v1.6.4) and [1.7.0](https://github.com/episource/unbound/tree/dev/module_ext_state/v1.7.0).

See also: https://www.nlnetlabs.nl/bugs-script/show_bug.cgi?id=1214
(Pull-Requests were disabled when I created this ticket)

PR #6, opened 2018-04-30
Python module: support multiple instances 

episource:
Currently, it's not possible to use multiple instances of the python module. Nevertheless, adding the word `python` to the `module-config:` option several times is not rejected by unbound. In the end, however, it's not a usable setup supported by unbound.

The proposed solution allows to reference the python module more than once within `module-config:` option. For each instance, a `python-script:` option needs to be added to the `python:` section, specifying the python script file to be used. The first python module picks up the first `python-script:` option, the second picks up the second script, and so on.

I've been successfully using this patch with unbound-[1.6.0](https://github.com/episource/unbound/tree/dev/pythonmod-multi-instance/v1.6.0), [1.6.1](https://github.com/episource/unbound/tree/dev/pythonmod-multi-instance/v1.6.1), [1.6.4](https://github.com/episource/unbound/tree/dev/pythonmod-multi-instance/v1.6.4) and [1.7.0](https://github.com/episource/unbound/tree/dev/pythonmod-multi-instance/v1.7.0).

See also: https://www.nlnetlabs.nl/bugs-script/show_bug.cgi?id=1213
(Pull-Requests were disabled when I created this ticket)

### DIFF
--- a/doc/unbound.conf.5.in
+++ b/doc/unbound.conf.5.in
@@ -1752,7 +1752,8 @@ clause gives the settings for the \fIpython\fR(1) script module.  This module
 acts like the iterator and validator modules do, on queries and answers.
 To enable the script module it has to be compiled into the daemon,
 and the word "python" has to be put in the \fBmodule\-config:\fR option
-(usually first, or between the validator and iterator).
+(usually first, or between the validator and iterator). Multiple instances of
+the python module are supported by adding the word "python" more than once.
 .LP
 If the \fBchroot:\fR option is enabled, you should make sure Python's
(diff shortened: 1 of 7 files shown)


episource:
If you are willing to accept all of the pull requests #4, #5, #6, you could also consider merging [dev/all-merged/master](https://github.com/episource/unbound/tree/dev/all-merged/master/) instead: All three patches have been merged into this single branch.

wcawijngaards:
Thank you for the python changes.  The pull requests did not exist yet when you created them.  We moved to github, so we can insert them from here.  I merged the all-merged branch as you suggested.

PR #7, opened 2018-10-19
allow query of "." without recursive flag

volth:
Resolving the root name is not usual, although some programs do this.
And they do it without setting recursive flag
So "dig +tra @127.0.0.1 example.com" (where unbound is on 127.0.0.1) fails because unbound rejects query of root name without  recursive flag.
It works with any other nameserver ("dig +tra @8.8.8.8 example.com", etc)

### DIFF
--- a/daemon/worker.c
+++ b/daemon/worker.c
@@ -1397,7 +1397,8 @@ worker_handle_request(struct comm_point* c, void* arg, int error,
 
 	/* If this request does not have the recursion bit set, verify
 	 * ACLs allow the snooping. */
-	if(!(LDNS_RD_WIRE(sldns_buffer_begin(c->buffer))) &&
+	if(qinfo.qname_len > 1 && /* non-recursive query "." is allowed */
+		!(LDNS_RD_WIRE(sldns_buffer_begin(c->buffer))) &&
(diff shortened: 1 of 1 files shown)


volth:
Of course this could be fixed by ```allow_snoop```, but it would be too permissive to allow all nonrecursive queries.

So I propose to make an exception only for "." which is in the unique position where there is no difference between nonrecursive and recursive

wcawijngaards:
Hi volth,

There is an important difference, that is that the TTL of the data can
be seen without altering cache contents.  And also data at . is
important, eg. root serial or root nameservers, or the root key.  And
this is what we want to stop, amongst other reasons, with the
nonrecursive block; the ability to probe the nameserver for the TTLs
without pulling in content to the cache, this is what is stopped.  And
also for .

If you really needed it it could become some sort of
allow_snoop_only_for_root ; but that would not become the default to
protect content at the . (root label domain name).

So, I don't see a way to do what you are asking for, and that seems to
make debug queries easier.  You know that the allow_snoop is possible to
set for particular IP addresses, it then overrides the global default
for every one, just have several access-control lines, the most specific
IP (netblock or IP-address) applies the setting to the incoming query.

Best regards, Wouter

PR #8, opened 2018-11-06
Fix OpenSSL without ENGINE support compilation

neheb:


wcawijngaards:
I think you want this:
    if (OPENSSL_VERSION_NUMBER < 0x10100000) && !defined(OPENSSL_NO_ENGINE)
because the call is deprecated at version 1.1.0 and before that only useful if not without ENGINE.

PR #9, opened 2018-11-06
Fix OpenSSL 1.0.2 compilation - deprecated APIs

neheb:
CRYPTO_set_id_callback got replaced by CRYPTO_THREADID_set_callback
which uses OpenSSL's threading API in 1.0.2. Made the necessary adjustments.


wcawijngaards:
Fixed, but with detection for that function, and backwards compatibility for older openssl versions.  That function is gone for openssl 1.1.0 by the way.

PR #10, opened 2018-11-11
build: fixed pkg-config operations

ensc:
PKG_PROG_PKG_CONFIG must not be in a conditional section because
autoconf expands this macro only once.  E.g. actually, builds without
python support will fail to detect systemd.

See in https://autotools.io/pkgconfig/pkg_check_modules.html the notes
at "3.4 Optional modules" for details.

Commit moves PKG_PROG_PKG_CONFIG out of the "if ..." condition.

Signed-off-by: Enrico Scholz <enrico.scholz@ensc.de>


wcawijngaards:
Thanks, applied.

PR #11, opened 2019-01-25
corrected the year of some recent changelog entries

jinmeiib:
(I believe this is trivial and doesn't require more explanation than the subject line; if not please ask).


wcawijngaards:
Hi Jinmei,

Thanks, applied! Sorry for the typos.

Best regards, Wouter

PR #13, opened 2019-02-06
Remove left-over requirements on OpenSSL >= 1.1.0 for cert name matching

dne:
When bug #4206 was resolved, some bits remained that still indicate that OpenSSL 1.1.0 or later is required:

- Invalid error logging when name verification is enabled with older OpenSSL (or LibreSSL).
- Outdated note in the unbound.conf(5) man page.


wcawijngaards:
Thanks, doc updated.

PR #17, opened 2019-04-08
A plugin to resolve records in mDNS via Avahi

janakj:
This is a pythonmodule based plugin for the Unbound resolver that can be used to resolve arbitrary DNS records in multicast DNS [RFC 6762] via Avahi. The plugin communicates with Avahi via DBus. The comment section at the beginning of the file contains detailed documentation.


wcawijngaards:
Thank you for the contribution!

PR #18, opened 2019-04-19
Fixes compile with srcdir overriden

new23d:
If not compiling within the source directory, overriding --srcdir to the configure script, make fails complaining about a missing source file. This patch includes the srcdir prefix to the file make complains about.


wcawijngaards:
This does not solve the issue correctly, the configlexer.c file is generated with flex from configlexer.lex in the srcdir.  And the generated file ends up in the build directory. Perhaps you do not have flex installed and that is why the build outside of srcdir fails?

new23d:
Not having flex installed was the issue indeed. Thank you for your feedback. I should've checked the configure output / raised an issue first.

wcawijngaards:
It seems to trip up other people too, the compile dependency on flex (and bison !).  Install bison for yacc, or you could be having the same problem with another source file.

PR #41, opened 2019-06-12
Feature/configure outbound msg retry

countsudoku:
I've added a option to configure the number of retries unbound will make in case of a THROWAWAY answer. In case of forward nameservers, this will restrict the number of retries for every nameserver (more precise for every target).

I try to use libunbound in a high performance application. Therefore it is important to restrict the number of request unbound will make.


wcawijngaards:
Hi Countsudoku,
This patch looks very good.  Currently in the process of fixing up bugs in the release so I do not want to introduce features (possible failures could be in it), right now.  So I won't accept the PR straight away, and leave it here for a bit to keep it separate from the release concerns.  Thanks for the excellent code!

countsudoku:
> Currently in the process of fixing up bugs in the release so I do not want to introduce features (possible failures could be in it), right now. So I won't accept the PR straight away, and leave it here for a bit to keep it separate from the release concerns.

Thanks for the info. You can also pick only some commits, since I've committed also some style fixes etc. I am not sure, if it would be better to re-generate the autogenerated files (util/configlexer.c, util/configparser.{h,c}) again with your flex version. Therefor I have committed these files in a separate commit. I am curious: Would it not be better to keep these autogenertaed files out of the repo and create them on demand?

countsudoku:
Is there any chance that this will be merged in the near future?

boarfish55:
Would love to see this merged as well. In the organization I work in, we sometimes have multiple unbound resolvers "chained" (one running on the client, then another one caching in front of the authorities), and the fact that they each retry 5 times by default on a THROWAWAY response means the latency at the client increases exponentially.

If we have a lame server on our network, or elsewhere, this means the client can sometimes have to wait 1+ second(s) before getting a reply, which in some cases causes timeouts. We would love to drop this value to one on "intermediate" resolvers and leave it to the client to make the decision as to retry to not.

wcawijngaards (approved):
The code looks good; have some smaller fixes I want to do. A lot of whitespace changes (of existing code), that is nicer in separate patches; diff that ignores whitespace can very quickly deal with the review of it. But it is okay here. I'll try to commit and fix the conflicts.

wcawijngaards:
Thanks for the patch! Merged it in the code repository.

Issue #52, opened 2019-07-31
New context object close()s old context’s debugout.


FGasper:
Related to https://github.com/NLnetLabs/unbound/issues/39

FGasper:
```
#include <stdio.h>
#include <unistd.h>
#include <unbound.h>

int main(void) {
    struct ub_ctx* ctx = ub_ctx_create();

    FILE *out_fh = fopen("./bug52-out", "w");
    if (!out_fh) {
        perror("fopen");
        return 1;
    }

    ub_ctx_debugout(ctx, out_fh);

    struct ub_ctx* ctx2 = ub_ctx_create();

    char str[] = "I am Iron Man!\n";

    if ( !fwrite(str, 1, sizeof(str), out_fh) ) {
        perror("fwrite");

        return 1;
    }

    return 0;
}
```
On my system:
```
gcc -L/opt/local/lib -I/opt/local/include -lunbound bug52.c -o bug52
```
… and the error:
```
> ./bug52
fwrite: Bad file descriptor
```

wcawijngaards:
See the PR #53 comments, another commit to fix it.  Let me know if that really fixes all the codepaths, apart from the reported.  Thanks for the report!

PR #53, opened 2019-07-31
Prevent a 2nd context from closing the 1st context’s log.

FGasper:
Issue #52

### DIFF
--- a/util/log.c
+++ b/util/log.c
@@ -102,10 +102,7 @@ log_init(const char* filename, int use_syslog, const char* chrootdir)
 		lock_quick_lock(&log_lock);
 	}
 	if(logfile && logfile != stderr) {
-		FILE* cl = logfile;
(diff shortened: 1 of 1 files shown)


wcawijngaards:
Your fix would stop all logfile closes.  That may work for you, but is not something for the generic code base.  The commit I just added creates a flag where it tracks if a debug file descriptor is in use, and when that is in use it tries to omit the file closures.  Also for the second context, the flag is a global variable.  I hope this fixes it.

wcawijngaards:
That did not work, but the second commit makes it work.  There was another log_init that needed to be stopped.  Does that also work for you, I mean, apart from the (very useful) example fail program?

FGasper:
@wcawijngaards My apologies for the non-responses here; with the time difference I think I just didn’t see the email notifications.

Thank you for looking into this and addressing the issues!

Issue #60, opened 2019-08-13
Add cache handling to libunbound


FGasper:
Would there be receptivity if I submitted PRs to augment libunbound with the ability to clear a specific query’s result from the caches?

I see logic like this in `daemon/remote.c`’s `do_cache_remove()` function.

What I’d like to be able to do is to do a query to see if a DNS propagation has taken effect, then redo that query in a few seconds if it hasn’t. Right now the only viable solution is recreating the libunbound context, which carries an undesirable overhead. Ideally, I’d be able to tell libunbound to clear just that specific query’s result, then redo the query.

ralphdolmans:
Sounds like useful functionality to have in libunbound. So yes, please submit a PR. Also, feel free to reach out if you have any question regarding the implementation.

PR #64, opened 2019-08-20
60 add cache handling

FGasper:
#60 

Pushing this up for feedback.

Thank you!


FGasper:
@ralphdolmans Right now the master branch fails `make test`:
```
util/log.c:120:30: error: use of undeclared identifier 'UB_SYSLOG_FACILITY'
                openlog(ident, LOG_NDELAY, UB_SYSLOG_FACILITY);
```
Is that something that should get its own issue, or is there some other/better way I can run the test suite against my branch?

Thank you!

wcawijngaards:
Hi Felipe, the master has updated and you have to re-run the configure to make this define available.  It then appears in config.h.  Just re-run ./configure to make the undeclared identifier error go away.
Best regards, Wouter

PR #72, opened 2019-09-06
SYSLOG_FACILITY custom option patch

ipv6labs:
Select LOCAL0 to LOCAL7 on configure.

--with-syslog-facility=LOCAL0 - LOCAL7

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -733,6 +733,9 @@
 /* default username */
 #undef UB_USERNAME
 
+/* SYSLOG_FACILITY */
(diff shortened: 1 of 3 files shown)


wcawijngaards:
Thanks for the patch, but we need a configure.ac patch for changes to configure.  I moved the changes to configure.ac, and committed the changes for `--with-syslog-facility`.

ipv6labs:
Thank you for the changes.
I'll be careful next time.

PR #76, opened 2019-09-17
Improvements and fixes for systemd unbound.service

Maryse47:
1. Remove `ProtectKernelTunables=true`: This prevents various with socket options from working as shown below.
`unbound[] warning: so-rcvbuf 1048576 was not granted. Got 425984. To fix: start with root permissions(linux) or sysctl bigger net.core.rmem_max(linux) or kern.ipc.maxsockbuf(bsd) values.`
2. Add `CAP_NET_ADMIN` to available caps which is needed for `ip-transparent: yes` config option to work as shown below.
`unbound[] warning: setsockopt(.. IP_TRANSPARENT ..) failed: Operation not permitted`
3. Make `ReadWritePaths` less permissive: `UNBOUND_SYSCONF_DIR` equals to `sysconfdir` which usually equals to `/etc` and `UNBOUND_LOCALSTATE_DIR` equals to `localstatedir` which usually equals to `/var`. Allowing write access for those dirs shouldn't be needed. The only dirs unbound should be allow to write to are `/run` ( for pidfile), `@UNBOUND_RUN_DIR@` (for chroot) and `@UNBOUND_CHROOT_DIR@` in case it differs from the previous one.
4. Bind-mount `/run/systemd/notify`, `UNBOUND_PIDFILE`, `/dev/log`, `/dev/urandom` in order to use them inside chroot.
5. Add few extra hardening options: `RestrictNamespaces`, `LockPersonality` and `RestrictSUIDSGID` should be safe to use.

@wcawijngaards


wcawijngaards:
Thanks for the improvements! Merged it.

PR #81, opened 2019-09-19
Consistently use /dev/urandom instead of /dev/random in scripts and docs

Maryse47:
Unbound code call /dev/urandom (see below)  but various docs and scripts mention /dev/random which may be confusing.

https://github.com/NLnetLabs/unbound/blob/release-1.9.3/compat/arc4random.c#L107
https://github.com/NLnetLabs/unbound/blob/release-1.9.3/compat/getentropy_linux.c#L251
https://github.com/NLnetLabs/unbound/blob/release-1.9.3/compat/getentropy_osx.c
https://github.com/NLnetLabs/unbound/blob/release-1.9.3/compat/getentropy_solaris.c#L116

@wcawijngaards

### DIFF
--- a/contrib/unbound.init
+++ b/contrib/unbound.init
@@ -54,10 +54,10 @@ start() {
 	[ -e ${rootdir}/dev/log ] || touch ${rootdir}/dev/log
 	mount --bind -n /dev/log ${rootdir}/dev/log >/dev/null 2>&1;
     fi;
-    if ! egrep -q '^/[^[:space:]]+[[:space:]]+'${rootdir}'/dev/random' /proc/mounts; then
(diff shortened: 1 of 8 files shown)


wcawijngaards:
Yes that changed in the implementation where open of dev happened at first in the util/random code, but then moved to the compat with contributed code because the random generators changed in libc('s in BSD and Linux), I guess this is where this could have changed.  Having a good supply is important because you do not want the stream of DNS replies cut short by a wait for randomness.  Which would be used for generating the 16-bit ID numbers in DNS packets.  Hence urandom, by the way.  So the changes you suggest are an improvement (in the comments around the code).  Thanks! Merged.

PR #82, opened 2019-09-20
Downgrade CAP_NET_ADMIN to CAP_NET_RAW in unbound.service

hardfalcon:
[Since kernel 3.2](https://github.com/torvalds/linux/commit/6cc7a765c2987f03ba278dac03c7cc759ee198e7), just using CAP_NET_RAW instead of CAP_NET_ADMIN is sufficient to allow for the usage of the IP_TRANSPARENT socket option. CAP_NET_ADMIN allows far more mayhem then CAP_NET_RAW, so prefer the safer, more restrictive solution.


wcawijngaards:
Thanks for the fixup, less permissions are better.  Merged it.

PR #83, opened 2019-09-20
unbound.service.in: do not fork into the background

Maryse47:
This is needed when unbound config doesn't set [do-daemonize: no](https://github.com/NLnetLabs/unbound/blob/release-1.9.3/doc/example.conf.in#L236) by itself otherwise starting service fails with:
` systemd[1]: unbound.service: Got notification message from PID <PID>, but reception only permitted for main PID which is currently not known`

@wcawijngaards


wcawijngaards:
Thanks!  Yes, that would make sure.  Merged.

Issue #84, opened 2019-09-20
systemctl reload fails


sam-lunt:
I'm running version 1.9.3-2 of unbound on Arch Linux.

The `CapabilityBoundingSet` value in `/usr/lib/systemd/system/unbound.service` causes the call `# systemctl reload unbound` to fail with the below error:

    kill[57844]: kill: sending signal to 57760 failed: Operation not permitted

If the line `CapabilityBoundingSet` is commented out in the service file, the call to reload works without issue. Also, if the value `CAP_KILL` is added to the `CapabilityBoundingSet`, the reload command is able to complete without error.

My guess is that the `ExecReload` command `/bin/kill -HUP $MAINPID` is executed as root. The combination of the fact that the unbound process is owned by the unbound user and that root does not have the `CAP_KILL` capability means that there is insufficient permissions for root to send the HUP signal to the unbound process.

wcawijngaards:
Fixed by merging #85.  Thanks for the fix.

PR #85, opened 2019-09-21
Add kill capability to systemd service file

sam-lunt:
As mentioned in [this issue](https://github.com/NLnetLabs/unbound/issues/84), the ExecReload command calls kills on a process owned by the unbound user (or whatever user is configured). To do so, it needs the CAP_KILL capability. I think adding this capability to the systemd unit makes sense, since it's using `kill` right there in the unit file. Additionally, since the process does not run as `root`, the extra capability will be dropped by the main process after start up


wcawijngaards:
Thanks!  Noticed that Maryse47 likes it and that was the contributor for the previous service file changes.  So it looks good, merged.

PR #86, opened 2019-09-24
Added -b / source address option to smallapp/unbound-anchor.c

psquarejho:
This is a patch which we need for the unbound server at a customer.

It allows you to specify the source address which unbound-anchor uses to connect to the root name servers. This is needed in our case as only the (additional) service ips on an interface are allowed to perform DNS queries, not the base ip of the interface.

Credit for developing this patch goes to Lukas Wunner <lukas@wunner.de>. I have just been maintaining this patch for the internal packages since 1.4, we just never got around to submit it upstream.

We hope to get this added to main stream unbound, as this is basically the last custom unbound patch we are maintaining.


wcawijngaards:
Thanks for the patch, that is a useful feature and nice code!  Merged.

PR #90, opened 2019-10-02
fix build with nettle-3.5

vcunat:
https://git.lysator.liu.se/nettle/nettle/commit/8bf4747d9


wcawijngaards:
Thanks for the build fix!  This works for me too (nettle 3.4.1).  Merged.

PR #95, opened 2019-10-19
[stub] add stub_tcp_upstream support

### LINKED ISSUE #94: stub-zone stub-tcp-upstream
Hi,

Looking at stub-zone config, it is possible to force TLS upstream with  `stub-tls-upstream`.

Would it make sense to also be able to configure `stub-tcp-upstream`?

Context, destination may block UDP (because they are battling reflection attack). They could rely on TCP to make sure the request they are getting are legitimate. Being able to force unbound to use TCP for specific stubs would help as currently, unbound seems to only fallback to TCP when receiving a response with TC set.

chantra:
This allows using TCP instead of UDP for a specific stub zone

Fixes #94

### DIFF
--- a/daemon/worker.c
+++ b/daemon/worker.c
@@ -1935,8 +1935,8 @@ worker_delete(struct worker* worker)
 struct outbound_entry*
 worker_send_query(struct query_info* qinfo, uint16_t flags, int dnssec,
 	int want_dnssec, int nocaps, struct sockaddr_storage* addr,
-	socklen_t addrlen, uint8_t* zone, size_t zonelen, int ssl_upstream,
-	char* tls_auth_name, struct module_qstate* q)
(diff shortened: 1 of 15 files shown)


chantra:
I did not update the smallapps as I was not compiling with `make` and those are skipped.
I would be happy to get those to compile if the idea is agreed upon.

chantra:
smallapps was only a couple of changes away, so I went ahead and applied changes.

ziollek:
@chantra - i've prepared [PR](https://github.com/NLnetLabs/unbound/pull/519) based on your changes which also allows using tcp upstreams in forward zones

pemensik:
Should the be closed, when PR #519 were merged?

chantra:
Thanks folks for pushing it through.

PR #97, opened 2019-10-22
manpage: Add missing word on unbound.conf

Erethon:
After having a quick glance on the RFC 5011 document, I'm assuming this is what is implied here.


wcawijngaards:
Thanks!  Merged.

PR #113, opened 2019-11-17
[unbound] force DoT to auth nameservers based on IP blocks

chantra:
This is the result of my IETF106 hackathon. It does work, but I did take
a few shortcuts here and there.

This diff adds a new configuration option `tls-auth-server`
which enabling telling to unbound that it should use DoT when talking to
the authoritative servers in the given subnets.

This is a "server" option with syntax:

tls-auth-server: <IP netblock> <name>

If <name> is empty string `""`, no verification will be performed.

example:
tls-auth-server: 2001:DB8:1234::/48 ""
tls-auth-server: 192.168.2.0/26 "somename.example.com"
tls-auth-server" 192.168.2.128 ""

This take advantage of the ACL logic and re-uses acl_list to store the
IP netblocks in and the associate hostname to verify TLS with.
Within iterator.c, and before sending the outgoing query, the target is
looked up in the acl_list, if there is a match, tls_auth_name and
ssl_upstream are set to True. The target's target->addr->sin_port is
also set to UNBOUND_DNS_OVER_TLS_PORT and the rest of the query will
follow its normal flow.

As a follow up, it could be good to rely on the same syntax
`forward-addr` and store that in the acl_list, so we can easily set
custom ports and piggyback on the current logic to parse TLS server
config options.


chantra:
cc @ralphdolmans

PR #122, opened 2019-11-25
In tcp_callback_writer(), don't disable time-out when changing to read.

he32:
I suspect this is the cause of the apparent non-working of the tcp-idle-timeout setting,
as observed in e.g.
 https://nlnetlabs.nl/pipermail/unbound-users/2019-August/011748.html


wcawijngaards (approved):
Looks good.  Must have been an oversight in the code changes for 1.9.0, I think.

wcawijngaards:
Thanks for the fix!  That is some serious deep code inspection, and what looks like a good bugfix to me.  I'll merge it in a bit and make a Changelog entry for it.

Issue #99, opened 2019-10-23
Memory leak in ub_ctx (event_base will never be freed)


WolverinDEV:
Hey,
the API method `ub_ctx_create_event`  combined with `ub_ctx_delete` leaks some memory (24 bytes).
This is because a new `event_base` will be allocated via `ub_libevent_event_base`, but never deallocated. Even if we call `ub_ctx_delete`.

wcawijngaards:
Thanks for the heads up!  Added a check that when the event base is set via this method it frees the struct that is alloced.

fobser:
This is only true for ub_libevent_event_base  in util/ub_event_pluggable.c.
The one in util/ub_event.c does not allocate memory but uses the passed in event_base.
So now I'm facing a use-after-free and most of the time event_dispatch() just returns.

WolverinDEV:
Well this is really an major issue. It has been implemented falsely.
Lookup this part of code:
https://github.com/NLnetLabs/unbound/blob/1d45b4a1e085acba51e27d6e7d360322959aaceb/libunbound/libunbound.c#L220-L225

I'll create a PR in a few min which will fix this flaw

ralphdolmans:
Hi Florian,

As far as I can tell libunbound always uses ub_event_pluggable.c, in which case this should not be an issue. That is assuming you build Unbound with the shipped Makefile. I see in the OpenBSD code that you are using your own Makefile which uses the ub_event.c, is that intentional?

fobser:
Edit: nevermind
I'm not sure, however, this is the knob in configure if I'm not mistaken:
  --enable-event-api      Enable (experimental) pluggable event base
                          libunbound API installed to unbound-event.h
It says experimental...

fobser:
unwind(8), which is the only consumer of libunbound in OpenBSD base, uses libevent so it felt natural to also stick to that.
I could investigate to switch to the plugable version if that's better / supported.

fobser:
I switched unwind over to ub_event_pluggable,c and it seems to be working.
Needs a whee bit more testing though.
In any case, I don't have a strong opinion in which direction this should go.
I'd suggest to delete ub_event.c though if that's deprecated / should not be used.

ralphdolmans:
> I'm not sure, however, this is the knob in configure if I'm not mistaken:
> --enable-event-api Enable (experimental) pluggable event base

[...]

PR #136, opened 2019-12-19
Fixing use-after-free issue which raised up while fixing issue #99

WolverinDEV:
Fixed the use-after-free issue within the libevent part of unbound.
This issue raised while attempting to fix issue #99.

### DIFF
--- a/libunbound/libunbound.c
+++ b/libunbound/libunbound.c
@@ -217,15 +217,22 @@ ub_ctx_create_event(struct event_base* eb)
 	ctx->created_bg = 0;
 	ctx->dothread = 1; /* the processing is in the same process,
 		makes ub_cancel and ub_ctx_delete do the right thing */
-	ctx->event_base = ub_libevent_event_base(eb);
+	if(eb) {
+		ctx->event_base = ub_libevent_event_base(eb);
+		ctx->event_base_malloced = 1;
+	} else {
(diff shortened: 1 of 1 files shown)


WolverinDEV:
Hold up wait a second.
I've messed the order up.

ralphdolmans:
Hi, thank you for this PR. Looks like the issue #99 fix is fine after all, see the comments on that issue.

I am also not sure if these changes would have worked if it was an issue. Looks like `eb` is always set, which would result is never calling free(), right?

fobser:
I wouldn't go as far as saying that the fix in #99 is fine after all.
It's is still a problem if one uses ub_event.
I just switched away from it, so *I'm* no longer effected by it.

WolverinDEV:
>  Looks like eb is always set, which would result is never calling free(), right?

I don't think so. You could call this method with an given base, which will then not be freed or with `NULL` which will result in an internal allocated ev base which will be freed.

ralphdolmans:
> You could call this method with an given base, which will then not be freed or with `NULL` which will result in an internal allocated ev base which will be freed.

But the `calloc` in `ub_libevent_event_base()` in `ub_event_pluggable.c` does not depend on the given base. When you give a base you will still have the memory allocation there. So the `free()` should happen when using `ub_event_pluggable.c`, which should be the case for libunbound, also when a base is provided.

PR #142, opened 2020-01-08
unbound.service.in: stop binding pidfile inside chroot dir

### LINKED ISSUE #138: systemd service fails on Ubuntu 18.04
unbound 1.9.6 fails to load as a systemd service on Ubuntu 18.04.3 LTS, when built from source and using the included `contrib/unbound.service` file.

unbound was built and installed using the following command line:
```bash
./configure --enable-systemd --with-libevent && make && make install && ldconfig
```

The following error message appears in `journalctl`:
```
unbound.service: Failed to set up mount namespacing: No such file or directory
unbound.service: Failed at step NAMESPACE spawning /usr/local/sbin/unbound: No such file or directory
unbound.service: Main process exited, code=exited, status=226/NAMESPACE
unbound.service: Failed with result 'exit-code'.
```

I narrowed it down to this line in `unbound.service`:
```
BindPaths=-/usr/local/etc/unbound/unbound.pid:/usr/local/etc/unbound/usr/local/etc/unbound/unbound.pid
```

I'm not sure why this causes `/usr/local/sbin/unbound` to not be found, but when I comment out the line, unbound works as expected. The line itself looks pretty strange, with the duplication of `/usr/local/etc/unbound`.

Maryse47:
Apparently pidfile isn't used inside chroot and binding it may cause some weird failures with older systemd.

Fixes https://github.com/NLnetLabs/unbound/issues/138

### DIFF
--- a/contrib/unbound.service.in
+++ b/contrib/unbound.service.in
@@ -26,7 +26,6 @@ ReadWritePaths=/run @UNBOUND_RUN_DIR@ @UNBOUND_CHROOT_DIR@
 TemporaryFileSystem=@UNBOUND_CHROOT_DIR@/dev:ro
 TemporaryFileSystem=@UNBOUND_CHROOT_DIR@/run:ro
 BindReadOnlyPaths=-/run/systemd/notify:@UNBOUND_CHROOT_DIR@/run/systemd/notify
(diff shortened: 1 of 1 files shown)


ralphdolmans:
thanks!

PR #147, opened 2020-01-21
minor #1344 change rfc reference for reserved top level dns names

mnacharov:
[RFC 7686](https://tools.ietf.org/html/) covers only `.onion` domain.

[RFC 2606](https://tools.ietf.org/html/rfc2606) covers four others:
* test
* example
* invalid
* localhost

Why `example` and `localhost` not set empty by default?

P.S.: issue link #1344 was taken from historical commit, where this comment was specified

### DIFF
--- a/doc/unbound.conf.5.in
+++ b/doc/unbound.conf.5.in
@@ -1296,7 +1296,7 @@ local\-data: "onion. 10800 IN
     SOA localhost. nobody.invalid. 1 3600 1200 604800 10800"
 .fi
 .TP 10
-\h'5'\fItest (RFC 2606)\fR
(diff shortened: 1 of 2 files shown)


ralphdolmans:
I think RFC6761 is the one that should be referenced. That document specifies that `example.` should be resolved normally.

`localhost.` should give back the loopback addresses when queried for address types, which Unbound does: https://github.com/NLnetLabs/unbound/blob/0feee99055786c24205970faacac6ed39835be28/services/localzone.c#L765

`test.` and `invalid.` should generate negative responses.

I will update the RFC references in the code to RFC6761.

ralphdolmans:
Thanks!

PR #151, opened 2020-01-24
Fixes for systemd units

Maryse47:
unbound.service.in: add RuntimeDirectory and ConfigurationDirectory

> Adding 'RuntimeDirectory' is needed when pidfile path is set to
> subdirectory under /run.
> 
> Adding ConfigurationDirectory may help in some non-standard setups.
> 
> Also add more descriptions about used rules to avoid user confusion
> about they meaning and purpose.

Move unbound_nochroot.service to unbound_portable.service

> The real purpose of this service is to make it work with
> https://systemd.io/PORTABLE_SERVICES/ which are incompatible with
> chroot workarounds from original unbound.service.
> 
> The service content is identical to unbound.service with exception
> for chroot related rules which were modified as needed.


wcawijngaards:
Yes, as I wrote https://github.com/NLnetLabs/unbound/pull/150#issuecomment-578661656 in a comment for #150, this looks like a good fix.  I'll wait a little more to hear your comments on the issue, perhaps it can be improved further.

Maryse47:
@wcawijngaards I left comment in https://github.com/NLnetLabs/unbound/pull/150#issuecomment-578718128 . We may find improvements for both generic and portable unit in the future but I think this one is good to go for now

Maryse47:
I rebased this to cover some new `unbound_nochroot` - `unbound_portable` transitions.

Maryse47:
Added `StateDirectory=unbound` as per https://github.com/NLnetLabs/unbound/pull/150#issuecomment-578728493

Maryse47:
@wcawijngaards I think all discussions were addressed now.

wcawijngaards (approved):
Review of changes looks good.  Thanks for the contributed patches!

PR #154, opened 2020-01-27
Allow use of libbsd functions with configure option --with-libbsd

edmonds:
Hi,

This pull request is for a patch developed by @stevenc99 and shipped in the Debian package of unbound ≥ 1.6.4. It adds a `--with-libbsd` configure option for using some functions shipped in the libbsd library rather than the implementations in unbound's `compat/` directory. We meant to submit this patch upstream back in 2017 but it looks like that never happened.

There are further details here:

• https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=853751

• https://wiki.debian.org/arc4random

I've refreshed the patch so that it is up-to-date with the Unbound master branch.

Thanks!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -881,6 +881,19 @@ fi
 fi
 AC_SUBST(SSLLIB)
 
+# libbsd
+AC_ARG_WITH([libbsd], AC_HELP_STRING([--with-libbsd], [Use portable libbsd functions]), [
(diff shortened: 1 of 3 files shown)


ralphdolmans:
Great, thanks!

Resolved merge conflict on libunbound.pc.in, from PR#155 changes.

PR #155, opened 2020-01-27
contrib/libunbound.pc.in: Fixes to Libs/Requires for crypto library dependencies

edmonds:
contrib/libunbound.pc.in: Only specify -lunbound for Libs

According to the pkg-config manpage, the "Libs" line in a .pc file should give the link flags "specific to your package", and specifically says not to include link flags for dependencies:

    Libs:  This line should give the link flags specific to your
           package.  Don't add any flags for required packages;
           pkg-config will add those automatically.

contrib/libunbound.pc.in: Embed the correct crypto dependencies

This commit removes the hardcoded dependency in the libunbound pkg-config .pc file on the libcrypto and libssl modules and instead populates the .pc file based on which crypto library was selected at configure time.

Note that the .pc file specifies pkg-config module names for the "Requires" line and this can vary from the library filename (e.g. "nss" is the pkg-config module name vs. "nss3" being the library name).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -761,6 +761,8 @@ AC_ARG_WITH([nss], AC_HELP_STRING([--with-nss=path],
 	fi
         LIBS="$LIBS -lnss3 -lnspr4"
 	SSLLIB=""
+	PC_CRYPTO_DEPENDENCY="nss nspr"
+	AC_SUBST(PC_CRYPTO_DEPENDENCY)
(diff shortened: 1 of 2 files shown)


wcawijngaards (approved):
These changes look good!

wcawijngaards:
Thanks for the PR, that makes it easier to use different crypto libraries.

PR #156, opened 2020-01-29
Added unbound-control view_local_datas_remove command

n3bul4:
I thought this would be a neat feature, to be able to bulk remove local_data from a view.


ralphdolmans (commented):
Thanks, the code looks good.
One thing; the new command should be also be added to unbound-control.8.in.

ralphdolmans (approved):
looks good, thanks

ralphdolmans:
Parentheses is unbound-control.c:761 are incorrect, fixed in https://github.com/NLnetLabs/unbound/commit/5fcae2f0e04f42a21138d9782c1c962b8e2edcf9

PR #159, opened 2020-02-03
Serve stale

gthess:


cachedb/cachedb.c
```diff
@@ -433,8 +428,14 @@ good_expiry_and_qinfo(struct module_qstate* qstate, struct sldns_buffer* buf)
...
 
+	/* Check if we are allowed to return expired entries:
+	 * - serve_expired needs to be set
+	 * - if SERVE_EXPIRED_TTL is set make sure that the record is not older
+	 *   than that. */
 	if((time_t)expiry < *qstate->env->now &&
-		!qstate->env->cfg->serve_expired)
+		(!qstate->env->cfg->serve_expired ||
+			(SERVE_EXPIRED_TTL &&
+			*qstate->env->now - (time_t)expiry > SERVE_EXPIRED_TTL)))
```
wcawijngaards, line 447: If `time_t` is unsigned, the now - expiry is going to be a large number, and thus give the wrong outcome. Perhaps rewrite to like now > constant + expiry that does not have the subtraction.  Or add a test that now > expiry, that could actually be in the if statement at the start, but I am not sure all cases have it.

wcawijngaards: Actually, it is safe, in the cases of the if-statement (in those cases the subtraction has a positive outcome), but on its own it would not be.

services/mesh.c
```diff
@@ -427,30 +557,27 @@ void mesh_new_client(struct mesh_area* mesh, struct query_info* qinfo,
...
+		goto servfail_mem;
 	}
 	if(rep->c->tcp_req_info) {
 		if(!tcp_req_info_add_meshstate(rep->c->tcp_req_info, mesh, s)) {
 			log_err("mesh_new_client: out of memory add tcpreqinfo");
 			goto servfail_mem;
 		}
 	}
+	/* add serve expired timer if required and not already there */
+	if(timeout && !mesh_serve_expired_init(s, timeout)) {
```
wcawijngaards, line 566: This exits the routine when out of memory, but does not print an error message.  Also inside mesh_serve_expired_init there is no error message (eg. for the regional_alloc call).

util/config_file.h
```diff
@@ -362,6 +362,11 @@ struct config_file {
 	int serve_expired_ttl;
 	/** reset serve expired TTL after failed update attempt */
 	int serve_expired_ttl_reset;
+	/** TTL for the serve expired replies */
+	int serve_expired_reply_ttl;
+	/** serve expired entries only after trying to update the entries and this
+	 *  timeout is reached */
```
wcawijngaards: add `in milliseconds`.  Also the serve_expired_reply_ttl and serve_expired_client_timeout do not have a recommended value shown (commented out) in doc/example.conf.in.

util/module.h
```diff
@@ -306,6 +306,17 @@ typedef int inplace_cb_edns_back_parsed_func_type(struct module_qstate* qstate,
...
 
+/**
+ * Function called when looking for (expired) cached answers during the serve
+ * expired logic.
+ * Called as func(qstate, lookup_qinfo)
+ * Where:
+ *	qstate: the query state.
+ *	lookup_qinfo: the qinfo to lookup for.
+ */
+typedef struct dns_msg* serve_expired_lookup_func_type(
```
wcawijngaards, line 317: It would be nice to have a fptr_wlist item for this func_type to check for the possible allowed values.

wcawijngaards (approved):
Looks good!  Thanks for looking into the zero TTL case.

services/mesh.c
gthess: Fixed in https://github.com/NLnetLabs/unbound/pull/159/commits/2a6274a33fedddb785099e55b88e3d8b22ab0788

util/config_file.h
gthess: Fixed in https://github.com/NLnetLabs/unbound/pull/159/commits/479b4c69e75ec22460263dcee159e9e749423273

util/module.h
gthess: Fixed in https://github.com/NLnetLabs/unbound/pull/159/commits/ea29698eebcfae31874d04f7e8d9564591d52dd2

PR #172, opened 2020-02-27
Add IBM s390x arch for testing

noloader:
This is a big-endian platform, so it is nice to have for insurance. The change [tested OK on my fork](https://github.com/noloader/unbound/runs/472974384).

This was not included in the original PR because I wanted to ensure the big-endian tests did not become a deal breaker.

This should be the last of my meddling with Unbound (for now :).


ralphdolmans:
Great, thanks!

PR #173, opened 2020-02-28
Update makedist.sh

noloader:
Use the latest config.guess and config.sub
Use SHA-256 for release signing

The last time I checked GPG - which was several years ago - GPG used SHA-1 as the default digest algorithm. I got into the habit of using `--digest-algo SHA256` to explicitly avoid SHA-1 in anticipation of [Announcing the first SHA1 collision ](https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html). (It was apparent the SHA-1 collision was coming based on Marc Stevens' work on [hashclash](https://github.com/cr-marcstevens/hashclash)).


wcawijngaards:
Thanks for the contribution.  Merged it.

Issue #175, opened 2020-02-29
.libs/unitmain.o:unitmain.c:function main: error: undefined reference to 'ENGINE_cleanup'


noloader:
Hi Everyone,

I'm testing Unbound on Android. Android testing is limited to compile and link. Also see [PR #174](https://github.com/NLnetLabs/unbound/pull/174).

OpenSSL and Expat are built for the tests. OpenSSL is built with `no-asm no-hw no-engine`.

The Android tests are failing with an error message from the linker:

```
.libs/unitmain.o:unitmain.c:function main: error: undefined reference to 'ENGINE_cleanup'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Makefile:360: recipe for target 'unittest' failed
```

In [`testcode/unitmain.c`](https://github.com/NLnetLabs/unbound/blob/master/testcode/unitmain.c), it looks like the [header is guarded as expected](https://github.com/NLnetLabs/unbound/blob/master/testcode/unitmain.c#L55):

```
#ifdef HAVE_OPENSSL_ENGINE_H
#include <openssl/engine.h>
#endif
```

However, the call to `ENGINE_cleanup()` [is not guarded](https://github.com/NLnetLabs/unbound/blob/master/testcode/unitmain.c#L920):

```
#  ifdef HAVE_OPENSSL_CONFIG
#  ifdef HAVE_EVP_CLEANUP
	EVP_cleanup();
#  endif
	ENGINE_cleanup();
	CONF_modules_free();
#  endif
```

PR #176, opened 2020-02-29
Fix link error when OpenSSL is configured with no-engine (GH #175)

noloader:
Also see Issue 175, [.libs/unitmain.o:unitmain.c:function main: error: undefined reference to 'ENGINE_cleanup'](https://github.com/NLnetLabs/unbound/issues/175).

I believe a new `configure` should be generated if this PR is accepted.


ralphdolmans:
Thanks! We already had a check for `ENGINE_cleanup` in daemon.c. I merged that check with the one you provided, for both daemon.c and unitmain.c

PR #180, opened 2020-03-03
Avoid calling exit in Travis script

noloader:
The Travis docs state calling exit should not be done. Blame Jeffrey Walton. Sorry about that.

Also see https://docs.travis-ci.com/user/job-lifecycle/ in the Travis docs.


wcawijngaards:
Thanks for the fix, that should help finish up the Travis builds better.

PR #182, opened 2020-03-05
Add iOS testing to Travis

noloader:
This PR adds iOS testing to Travis.

The iOS testing is similar to Android testing. The environment is set, OpenSSL and Expat are built, and then Unbound is built. iOS uses a prefix of `$HOME/iPhoneOS-$IOS_CPU`, `$HOME/AppleTVOS` or `$HOME/WatchOS`.

Android testing was changed to use a prefix of `$HOME/android$ANDROID_API-$ANDROID_CPU` to make the procedure consistent with iOS. It had the added benefit we did not need to expalin problems with installing into sysroot.

iOS builds for AppleTVOS are currently failing. I don't have a reason why at the moment. I've got another project running a similar configuration. For the other project, AppleTVOS builds are OK, but WatchOS builds are failing. Since TV and Watch seem to have chronic problems they are listed in `allow_failures`. (For the other project and failures see [here](https://travis-ci.org/noloader/cryptopp/builds/658210859)).

Both Android and iOS now perform a `make install` to test installation into prefix.

`sudo: false` was also removed. Current Travis machinery ignore the setting and issue a warning when the `sudo` key is set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: c
-sudo: false
 
 git:
   depth: 5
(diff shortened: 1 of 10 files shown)


wcawijngaards (approved):
Changes look good

wcawijngaards:
Thanks for the fixes, I appreciate the further fixups to the battery of scripts.

PR #191, opened 2020-03-11
Update iOS testing on Travis

noloader:
This PR updates Travis testing for OS X and iOS.

The two big changes were updating `setenv-ios.sh` to use a specific `-mxxx-min-ver`, which allowed iSO simulator testing. Without `-mxxx-min-ver` changes the simulators would fail to compile under current versions of Xcode.

* Update Brew packages in `.travis.yml`
* Add `TEST_OSX=yes` to differentiate between OS X and iOS in `.travis.yml`
* Use more consistent `name:` in `.travis.yml`
* Switch to `IOS_PREFIX="$HOME/$IOS_SDK-$IOS_CPU"` due to sdk/cpu combos
* Add iPhoneSimulator, AppleTVSimulator and WatchSimulator testing
* Install Android and iOS tools using `before_script` in `.travis.yml`
* Remove `"$CFLAGS"` from `install_openssl.sh`. OpenSSL already includes them.
* Add `no-tests` and `no-unit-tests` to `install_openssl.sh`. OpenSSL was calling `fork` in several places
* Patch OpenSSL in `install_openssl.sh`. OpenSSL was still calling `fork` in several places
* Use `pkg-config` name rather than `pkgconfig`. It was causing a failure in `apt-get`
* Update `15-ios.conf` for OpenSSL. Flags are now added in the base recipe
* Update `setenv_ios.sh`. It avoids `-mios-version-min` used in earlier Xcodes. It now uses a specific flag, like `-miphoneos-version-min`, `-miphonesimulator-version-min`, `-mappletvos-version-min`, `-mappletvsimulator-version-min`, `-mwatchos-version-min` and `-mwatchsimulator-version-min`
* Use `XCODE_` prefix consistently in `setenv_ios.sh`
* Guard use of `fork` and `execlp` in `daemon/unbound.c`, `libunbound/libworker.c` and `smallapp/unbound-control.c`

Travis is kind of slow today. You can see the result of the iOS changes at [Job 152871267](https://travis-ci.com/github/noloader/unbound/builds/152871267).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,12 @@ addons:
     - libevent-dev
     - libexpat-dev
     - clang
+  homebrew:
+    packages:
+      - openssl
+      - libevent
+      - expat
+    update: true
 
 jobs:
   include:
@@ -25,6 +31,7 @@ jobs:
       name: Clang on OS X, Amd64
       compiler: clang
       arch: amd64
+      env: TEST_OSX=yes
     - os: linux
       name: UBsan, GCC on Linux, Amd64
       compiler: gcc
@@ -81,7 +88,7 @@ jobs:
       dist: bionic
     - os: osx
       osx_image: xcode10
-      name: Apple iPhone, OS X, Amd64
+      name: Apple iPhone on iOS, armv7
       compiler: clang
       env:
         - TEST_IOS=yes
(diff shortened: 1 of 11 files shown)


ralphdolmans:
Thanks!

PR #194, opened 2020-03-17
Add libevent testing to Travis

noloader:
This PR adds libevent testing to Travis.

* Add three Travis jobs for GCC on Linux, Clang on Linux and Clang on OS X
* Add Homebrew recipe to install OpenSSL and libevent on OS X
* Fix `pkg-config` package name
* Refactor Travis `jobs:` to use `CONFIG_OPTS`
* Refactor Android tools install into `before_script:`
* Some whitespace cleanup in `configure.ac`
* Add libevent testing feedback to `unitmain.c`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,11 @@ addons:
     - libevent-dev
     - libexpat-dev
     - clang
+  homebrew:
+    packages:
+      - openssl
+      - libevent
+    update: true
 
 jobs:
   include:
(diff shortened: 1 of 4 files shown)


ralphdolmans:
Thanks again! This had a merge conflict in .travis.yml with the changes from PR#191, these should now be resolved. Travis is really slow right now, so we will need a bit of patience to see if it all works.

noloader:
No problems @ralphdolmans. Thanks.

PR #200, opened 2020-03-23
add ip-dscp option to specify the DSCP tag for outgoing packets

yrk-lab:
# Summary

DSCP, the Differentiated Services codepoint, is a field in the IP header packets used for QoS. It replaces the obsolete TOS field in IPv4 and the Traffic Class field in IPv6, per _RFC 3260_.

Let's add _ip-dscp_ server option to specify an integer value to be passed to `setsockopt(…, IP_TOS, …)` for every internet-family socket created, effectively enabling the QoS tagging of all outgoing packets.

# Testing

Parsing of the new option confirmed with _unbound-checkconf_ on `testdata/04-checkconf.tdir/good.all` and `testdata/04-checkconf.tdir/bad.dscp`.

A back-port to 1.9.6 was baked on production systems; the fact the outgoing traffic gets tagged is confirmed with tcpdump and other network monitoring tools:

```
11:09:31.574171 IP6 (class 0x8c, hlim 127, next-header UDP (17) payload length: 97) …
11:09:32.004307 IP (tos 0x8c, ttl 96, id 24361, offset 0, flags [none], proto UDP (17), length 79) …
```

cc @ralphdolmans @chantra


wcawijngaards (approved):
Changes look good, I'll make some modifications after merge.  What is it useful for to set the DSCP tag on DNS traffic?

wcawijngaards:
Did you really mean with the case fallthrought to set both IPv6 and IPv4 TOS on IPv6 sockets?  I now fixed it to set IPv6 for IPv6 sockets and IPv4 for IPv4 sockets.  But the initial code has a case statement fallthrough which makes IPv4 TOS for IPv6 sockets in addition to the IPv6 socket TCLASS.

wcawijngaards:
Thanks for the contribution!  Merged it and some code alterations.  Please let me know about the fallthrough so we can be sure to set the right socket options.  I am also curious what it is used for.

yrk-lab:
@wcawijngaards Thanks for accepting and the follow-up fixes!

It is useable in QoS settings to differentiate the control-plane traffic like DNS and ensure is not pushed off the queues by the regular data traffic.

As for the fall though, thanks for the catch: it went under the radar. Let me verify if it works the same way with the breaks in place.

yrk-lab:
@wcawijngaards Back-ported the fall-through fix – confirmed to work as intended. Thanks!

wcawijngaards:
Cool that you tested it, that is very nice.  Another bug removed by code review :-)

PR #201, opened 2020-03-25
Fix OpenSSL cross-compaile warnings

noloader:
This PR cleans up OpenSSL warnings when cross-compiling. It also uses the same `15-android.conf` and `setenv-android.sh` as Unbound.

OpenSSL has a penchant for ignoring user flags. When Travis jobs were added that used `ANDROID_API=23` (the minimum) it produced warnings like the following because OpenSSL stomped on Unbound's flags.

```
armv7a-linux-androideabi23-clang  -I. -Iinclude -fPIC -pthread -D__ANDROID_API__
=23 -march=armv7-a -mthumb -mfloat-abi=softfp -funwind-tables -fexceptions --sys
root=/home/travis/android-ndk/toolchains/llvm/prebuilt/linux-x86_64/sysroot -DOP
ENSSL_USE_NODELETE -DOPENSSL_PIC -DOPENSSLDIR="\"/home/travis/android23-armv7a\"
" -DENGINESDIR="\"/home/travis/android23-armv7a/lib/engines-1.1\"" -D__ANDROID_A
PI__=29 -DNDEBUG  -MMD -MF apps/app_rand.d.tmp -MT apps/app_rand.o -c -o apps/ap
p_rand.o apps/app_rand.c

In file included from <built-in>:374:
<command line>:6:9: warning: '__ANDROID_API__' macro redefined
      [-Wmacro-redefined]
#define __ANDROID_API__ 29
        ^
<command line>:1:9: note: previous definition is here
#define __ANDROID_API__ 23
        ^
```

The issue is cleared by removing the function `android_ndk` from `15-android.conf`.

A sample build with the noisy warnings is available [here](https://travis-ci.org/github/noloader/unbound/jobs/666678999)

### DIFF
--- a/contrib/android/setenv_android.sh
+++ b/contrib/android/setenv_android.sh
@@ -1,16 +1,43 @@
 #!/usr/bin/env bash
 
-# Error checking
+# ====================================================================
+# Sets the cross compile environment for Android
+#
+# Based upon OpenSSL's setenv-android.sh by TH, JW, and SM.
+# Heavily modified by JWW for Crypto++.
+# Updated by Skycoder42 for current recommendations for Android.
+# Modified by JWW for Unbound.
+# ====================================================================
+
+#########################################
+#####        Some validation        #####
+#########################################
+
+if [ -z "$ANDROID_API" ]; then
+    echo "ANDROID_API is not set. Please set it"
+    [[ "$0" = "${BASH_SOURCE[0]}" ]] && exit 1 || return 1
+fi
+
+if [ -z "$ANDROID_CPU" ]; then
+    echo "ANDROID_CPU is not set. Please set it"
+    [[ "$0" = "${BASH_SOURCE[0]}" ]] && exit 1 || return 1
+fi
+
 if [ ! -d "$ANDROID_NDK_ROOT" ]; then
     echo "ERROR: ANDROID_NDK_ROOT is not a valid path. Please set it."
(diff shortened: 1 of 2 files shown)


contrib/android/setenv_android.sh
```diff
@@ -65,7 +84,7 @@ case "$AOSP_CPU" in
     CXXFLAGS="-march=armv7-a -mthumb -mfloat-abi=softfp -funwind-tables -fexceptions -frtti"
     ;;
 
-  armv8|armv8a|aarch64|arm64|arm64-v8a)
+  armv8*|aarch64|arm64)
```
gthess, line 87: Was it supposed to be `arm64*` here or the change is indeed intended?
The `arm64-v8a` option is not accounted for anymore but I also see that we don't use it with travis.

noloader: I believe Android uses `arm64` alone. But `arm64*` should be fine too. `arm64*` will help users who do something like `ANDROID_CPU=arm64-v8a`. I'd make the change if I were you.

gthess: :+1:

noloader: @gthess, I grabbed the change for Crypto++ at [Commit ccb3e869eddd](https://github.com/weidai11/cryptopp/commit/ccb3e869eddd). Thanks.

PR #207, opened 2020-03-30
Clarify if-automatic listens on 0.0.0.0 and ::

wtoorop:
Current manpage documentation caused some confusion among DNS operators which was discussed in the freenode #dns channel.


wcawijngaards:
Thanks for the fixup in the documentation :-)

PR #208, opened 2020-03-30
Fix uncached CLIENT_RESPONSE'es on stateful transports

wtoorop:
I noticed dnstap CLIENT_RESPONSE did not contain the response sent to the client when:
- the client is connected on a stateful transport, **and**
- the response did not come from cache.

When investigating the issue, I noticed that repinfo->c->buffer did not contain the response in this case (second half of comm_point_send_reply).
Only after tcp_req_info_send_reply() was called, the the buffer contains the response.

Is it safe to change this order?

### DIFF
--- a/util/netevent.c
+++ b/util/netevent.c
@@ -3153,18 +3153,21 @@ comm_point_send_reply(struct comm_reply *repinfo)
 			&repinfo->addr, repinfo->c->type, repinfo->c->buffer);
 #endif
 	} else {
+#ifdef USE_DNSTAP
+		if(repinfo->c->tcp_parent->dtenv != NULL &&
+		   repinfo->c->tcp_parent->dtenv->log_client_response_messages)
+			dt_msg_send_client_response(repinfo->c->tcp_parent->dtenv,
+			&repinfo->addr, repinfo->c->type,
+			( repinfo->c->tcp_req_info
+			? repinfo->c->tcp_req_info->spool_buffer
+			: repinfo->c->buffer ));
+#endif
 		if(repinfo->c->tcp_req_info) {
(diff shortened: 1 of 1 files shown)


wcawijngaards:
No, it may be somewhere else, in a queue.

wcawijngaards:
I think this may be better.  Also buffer may no longer be there, because allocated in queue for tcp responses.
```diff --git a/util/netevent.c b/util/netevent.c
index 09023838..049c7ff0 100644
--- a/util/netevent.c
+++ b/util/netevent.c
@@ -3157,7 +3157,7 @@ comm_point_send_reply(struct comm_reply *repinfo)
                if(repinfo->c->tcp_parent->dtenv != NULL &&
                   repinfo->c->tcp_parent->dtenv->log_client_response_messages)
                        dt_msg_send_client_response(repinfo->c->tcp_parent->dtenv,
-                       &repinfo->addr, repinfo->c->type, repinfo->c->buffer);
+                       &repinfo->addr, repinfo->c->type, (repinfo->c->tcp_req_info?repinfo->c->tcp_req_info->spool_buffer:repinfo->c->buffer));
 #endif
                if(repinfo->c->tcp_req_info) {
                        tcp_req_info_send_reply(repinfo->c->tcp_req_info);
```

wtoorop:
Yes that works! Thanks! I've done a new commit with your proposed fix.

wcawijngaards:
Okay let's merge it, then.

wtoorop:
Do you want me to push the merge button?

wcawijngaards:
Sure! I don't know who should merge, but I just did.

Good to have this fixed.

PR #211, opened 2020-04-02
Down- and upstream padding a la RFC7830 & RFC8467

wtoorop:
When padding needs to be done, an EDNS0 Padding option is appended to edns_data's opt_list, but it is then still zero sized.  Padding block size is carried in a new data member in edns_data (padding_block_size) and only applied at the very end when encoding the packet.


wcawijngaards (approved):
Sure that looks ok with me.

wcawijngaards:
I would prefer the default for `pad-queries:` as `yes` instead of off right now.  Because that is a secure default.  Even if breaks stuff (does it?).

wtoorop:
It should not break stuff and is also only applied on TLS connections (which I assume support all support edns0 fine), so yes, I suppose this would be safe.
One more commit will follow then.

skoef:
I wanted to try this patch and apply it against master. Since it has merge conflicts and the generated natured of some of the affected files, I chose to remove the changes to `util/configlexer.c`, `util/configparser.c` and `util/configparser.h`. This made it necessary to add the following change to `util/configparser.y`:

```diff
@@ -263,6 +263,8 @@
 	server_shm_enable | server_shm_key | server_fake_sha1 |
 	server_hide_trustanchor | server_trust_anchor_signaling |
 	server_root_key_sentinel |
+	server_pad_responses | server_pad_responses_block_size |
+	server_pad_queries | server_pad_queries_block_size |
 	server_ipsecmod_enabled | server_ipsecmod_hook |
 	server_ipsecmod_ignore_bogus | server_ipsecmod_max_ttl |
 	server_ipsecmod_whitelist | server_ipsecmod_strict |
```

otherwise the new keywords would not be considered valid during configuration check. Perhaps that is solely needed if you don't apply the changes to mentioned files, but I got warnings like `warning: nonterminal useless in grammar: server_pad_responses`.

skoef:
In this patch, padding is only applied when the connection is both over SSL and TCP. On my setup, the unbound service would be behind a loadbalancer (dnsdist in my case) and the loadbalancer doesn't apply the padding itself. Unbound, with this patch doesn't either since the connection between dnsdist and unbound is not encrypted (but on a trusted network) and over UDP. 

Neither RFC7840 nor RFC8467 limits the application of padding to SSL + TCP connections (correct me if I'm wrong!) so would it perhaps make sense to drop that limitation?

wtoorop:
@skoef Thanks for the heads-up on the missing keywords in `configparser.y`!
@skoef I am reluctant about enabling padding for other transports than DoT, though I do acknowledge your use case. Are running on Linux? I think this might be resolvable with XDP... @DRiKE , WDYT?

util/config_file.h
```diff
@@ -600,6 +600,17 @@ struct config_file {
 	size_t dnscrypt_nonce_cache_size;
 	/** number of slabs for dnscrypt nonces cache */
 	size_t dnscrypt_nonce_cache_slabs;
+
+	/** EDNS padding according to FC7830 and RFC8467 */
```
gthess: Typo: FC -> RFC

services/outside_network.c
```diff
@@ -3026,7 +3037,9 @@ outnet_serviced_query(struct outside_network* outnet,
 		sq = serviced_create(outnet, buff, dnssec, want_dnssec, nocaps,
 			tcp_upstream, ssl_upstream, tls_auth_name, addr,
 			addrlen, zone, zonelen, (int)qinfo->qtype,
-			qstate->edns_opts_back_out);
+			qstate->edns_opts_back_out,
+			( ssl_upstream && env->cfg->pad_queries
```
gthess, line 3041: nit: extra space after '('

util/config_file.h
```diff
@@ -600,6 +600,17 @@ struct config_file {
...
+	/** true to enable padding of responses (default: on) */
+	int pad_responses;
+	/** block size with which to pad encrypted responses (default: 468) */
+	size_t pad_responses_block_size;
+	/** true to enable padding of queries (default: off) */
```
gthess: From the code, I think it should read "default: on"

gthess (approved):
Looks good to me too; some nits only. Compiled and run here with no problems.

It could be also good to have test cases for the padding but I don't know how, easily.
We have small programs that can be used for queries over TLS (e.g., testcode/streamtcp for the client side) that padding could be added as an option.

util/config_file.h
wtoorop: I find it more readable that way :)

PR #214, opened 2020-04-07
unbound-control-setup recreate certificates

gearnode:
The same patch than the one I did on nsd.

ref: https://github.com/NLnetLabs/nsd/pull/91


wcawijngaards (approved):
Ok, this is good to go.  Thanks for contributing it!

wcawijngaards:
Oops wrong button, let me merge that again.

PR #225, opened 2020-04-21
KSK-2010 has been revoked

akhait:
KSK-2010 is already revoked by ICANN: https://www.icann.org/news/blog/icann-is-revoking-the-old-key-signing-key-this-week


wcawijngaards:
Thanks!  That should update our list of trust anchors to the latest.

PR #234, opened 2020-05-10
Ensure proper alignment of cmsg buffers

jcourreges:
The cmsg macros expect a control message buffer to be aligned like
a struct cmsghdr.  The current layout around those stack-allocated
buffers probably provides the required alignment (usually 4 bytes).
Use a union to enforce proper alignment, in case future changes modify
the stack layout.

Spotted when chasing an unrelated bug with Otto Moerbeek (@omoerbeek).


ralphdolmans:
Thanks. This in need makes sense to do to prevent a potential future issue.

There are some `sizeof(control)`s left in log_assert statements in  `comm_point_send_udp_msg_if`, these also need to be changed to `control.buff`. I'll make these changes and merge this.

PR #268, opened 2020-07-20
draft-ietf-dnsop-serve-stale-10 has become RFC 8767 on March 2020

and0x000:
update documentation


gthess:
Thanks, we are in progress of releasing 1.11.0, so this will be merged right after that.

PR #269, opened 2020-07-22
Fix python module len() implementations

tobbez:
They were all missing the `self.` prefix when accessing the `obj`
instance variable, causing the following exception when attempting to
call `len()` on (for example) a `ReplyInfo_RRSet`:

```
  File "/usr/lib/python3.7/site-packages/unboundmodule.py", line 377, in __len__
    def __len__(self): return obj.rrset_count
NameError: name 'obj' is not defined
```


ralphdolmans:
Thanks! I'll merge this after the 1.11 release.

PR #270, opened 2020-07-23
munin plugin: always exit 0 in autoconf

cgzones:
The autoconf operation should always exit 0, also in case the answer in "no",
see https://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-concise.html#autoconf


wcawijngaards (approved):
Thanks!  This looks good, and can be merged after the 1.11.0 release that is in progress.

PR #272, opened 2020-07-27
Add EDNS client tag functionality

ralphdolmans:
Add option to include 16 bit EDNS tags in upstream queries, for specified destination addresses. Max one tag per query, the most specific match is selected.

Tags are configured using the `edns-client-tag: <IP netblock> <tag data>` configuration option, which is allowed multiple times. Tag data is configured in decimal format.


util/edns.h
```diff
@@ -42,11 +42,66 @@
...
+	uint16_t tag_data;
+};
+
+/**
+ * Create structure to hold EDNS tags
+ * @return: newly created edns_tags, NULL on alloc failure.
+ */
+struct edns_tags* edns_tags_create(void);
+
+/** Delete ENDS tags structure
```
wcawijngaards: s/ENDS/EDNS/

util/module.h
```diff
@@ -520,6 +520,7 @@ struct module_env {
 	struct edns_known_option* edns_known_options;
 	/* Number of known edns options */
 	size_t edns_known_options_num;
+	struct edns_tags* edns_tags;
```
wcawijngaards, line 524: A comment line above the member, like the other entries.  `/** EDNS client tag information */` or so.

util/edns.h
ralphdolmans: thanks, fixed

util/module.h
ralphdolmans: thanks, fixed

wcawijngaards (approved):
Review looks fine for me.

PR #275, opened 2020-07-30
Add feature to return the original instead of a decrementing TTL

rijswijk:
This PR adds a feature called **serve-original-ttl** which, if enabled, will make Unbound return the original TTL for records as received from upstream. This feature may be useful if Unbound serves as a front-end to, e.g., a hidden authoritative name server.

Notes on the PR:

- The caching logic was not changed, the TTL values in `packed_rrset_data` structures are treated the same as before. Instead, the new code stores an additional value in this structure call `ttl_add` which records how much has been added to the TTLs in the structure, and can be subtracted from these absolute TTLs instead of the current time when encoding responses.
- The logic in `packed_rrset_encode` in `msgencode.c` was changed slightly to make the time adjustment explicit. The original code sometimes modified the value of the function's input parameter `timenow`, setting it to 0. With the introduction of the new feature, this would have led to messy conditionals when adjusting the TTLs during encoding.
- Based on feedback from @wcawijngaards on an earlier version of this code the logic when decrementing TTLs in `cachedb/cachedb.c` was also changed such that the semantics of the decrement should be preserved while the original TTL will still be returned if this feature is enabled.
- The manual page explicitly specifies that enabling this feature means Unbound will ignore **cache-min-ttl** and **cache-max-ttl**.
- When allocating space for `struct packed_rrset_data` the code was changed in several places to explicitly initialise the newly allocated memory to zero, using either `regional_alloc_zero` or `calloc(1, <size>)`.


util/data/msgreply.c
```diff
@@ -197,9 +199,9 @@ rdata_copy(sldns_buffer* pkt, struct packed_rrset_data* data, uint8_t* to,
 		if(*rr_ttl > MAX_NEG_TTL)
 			*rr_ttl = MAX_NEG_TTL;
 	}
-	if(*rr_ttl < MIN_TTL)
+	if(!SERVE_ORIGINAL_TTL && (*rr_ttl < MIN_TTL))
```
ralphdolmans, line 202: Should a similar check be added to adjust_ttl() in val_sigcrypt.c?

util/data/msgencode.c
```diff
@@ -479,11 +483,10 @@ packed_rrset_encode(struct ub_packed_rrset_key* key, sldns_buffer* pkt,
 				return r;
 			sldns_buffer_write(pkt, &key->rk.type, 2);
 			sldns_buffer_write(pkt, &key->rk.rrset_class, 2);
-			if(data->rr_ttl[j] < timenow)
+			if(data->rr_ttl[j] < adjust)
```
ralphdolmans, line 486: Is this intentionally inconsistent with the checks in packed_rrset.c (where the comparison is still `d->rr_ttl[i] < now`)?

util/data/packed_rrset.c
```diff
@@ -353,11 +354,12 @@ packed_rrset_copy_region(struct ub_packed_rrset_key* key,
...
 		if(d->rr_ttl[i] < now)
 			d->rr_ttl[i] = SERVE_EXPIRED?SERVE_EXPIRED_REPLY_TTL:0;
-		else	d->rr_ttl[i] -= now;
+		else	d->rr_ttl[i] -= SERVE_ORIGINAL_TTL ? data->ttl_add : now;
 	}
 	if(d->ttl < now)
 		d->ttl = SERVE_EXPIRED?SERVE_EXPIRED_REPLY_TTL:0;
-	else	d->ttl -= now;
+	else	d->ttl -= SERVE_ORIGINAL_TTL ? data->ttl_add : now;
+	d->ttl_add = 0; /* TTLs have been made relative */
```
ralphdolmans, line 367: For my understanding, is the `d->ttl_add = 0` here "just to be safe", or really needed? Guess it is fine to have it either way.

ralphdolmans (approved):
Looks good! I have added some comments.

It would be nice to also have a testbound test that makes use of this option.

util/data/msgreply.c
rijswijk: Yes, possibly, although this is a bit tricky. The upward adjustment for `MIN_TTL` is fixable, but the downward adjustment if the expiration time comes before the TTL expires is maybe not something to change, what do you think?

util/data/msgencode.c
rijswijk: The reasoning was that here the original code performed an adjustment depending on the flag `PACKED_RRSET_FIXEDTTL` being set. This adjustment changed an input parameter to the function (`timenow`). I made this explicit, because the change for the new feature would add additional decision logic here, and it felt wrong to use an input variable rather than something explicit. If you would prefer this to be explicit everywhere, I can also change the code in `packed_rrset.c`, let me know.

util/data/packed_rrset.c
rijswijk: Since I don't know what will happen to the data in `struct packed_rrset_data` I added this just to be safe.

util/data/msgreply.c
wcawijngaards: I think it is best to have dnssec expiration remove the cache entry; so it can be refreshed with a hopefully correct value.  Otherwise the server would keep serving a dnssec invalid value.  So no change in val_sigcrypt is needed.

util/data/msgencode.c
wcawijngaards: Yes I would like you to also do it like that in packed_rrset.c, because I want to make sure that it does not subtract lower than 0, so the compare < adjust is better than < now to make really sure.  Because otherwise it could create negative, high-bit-set TTLs, and those are considered a resolver problem.

wcawijngaards (approved):
Looks good, with comment on comparison in packed_rrset.c improvement.

util/data/msgencode.c
rijswijk: I addressed this in commit d253db0, can you check if that meets your expectations?

wcawijngaards: Okay from me!

gthess (approved):
This is also a go from me!
I'll add a testcase for this and merge it. Is this OK @rijswijk?

rijswijk:
Yup @gthess go ahead, thanks!

Issue #279, opened 2020-08-03
macOS is broken? HAVE_FORK is undef


FGasper:
20dbd6339
^^ This commit appears to have added logic that breaks the following CI run:
https://travis-ci.org/github/FGasper/p5-DNS-Unbound/jobs/714547957

It appears that HAVE_FORK is being undef’d because this code:
```
#if defined(TARGET_OS_TV) || defined(TARGET_OS_WATCH)
#undef HAVE_FORK
#endif
```
… should check truthiness as well as defined-ness.

FGasper:
For context, those constants are defined thus on my machine:
```
        #define TARGET_OS_OSX               1
        #define TARGET_OS_IPHONE            0
        #define TARGET_OS_IOS               0
        #define TARGET_OS_WATCH             0
        
        #define TARGET_OS_TV                0
```

PR #280, opened 2020-08-03
Make tvOS & watchOS checks verify truthiness as well as definedness.

FGasper:
This fixes issue #279.


gthess:
Thanks for this, merged!

PR #284, opened 2020-08-05
DLV removal

### LINKED ISSUE #246: Remove DLV entirely from Unbound
Hey, I could be mistaken, but it seems like you might still have the DLV option in Unbound. If you do, it is time to remove it.

1.5.4 changed the default settings and removed the ISC DLV key from the packages - which is great, but at this point, any resolver that is querying the DLV is going extra work for zero added value. At this point there is an empty zone at dlv.isc.org, just so that resolvers querying it won't get stuck in a loop re-querying it.  It would be better to remove the DLV from Unbound entirely.  We removed it from BIND as of BIND 9.16 (the 2020 stable version), which has the added benefit of simplifying our code.

wcawijngaards:
This set of commits removes DLV support from Unbound.  DLV has been decommissioned and we advised to stop using it, now it is removed.  The config options for it print a warning.


gthess (approved):
Looks good to me. I especially like the warnings for the configuration options.

util/configparser.y
```diff
@@ -1120,15 +1120,13 @@ server_root_hints: VAR_ROOT_HINTS STRING_ARG
 server_dlv_anchor_file: VAR_DLV_ANCHOR_FILE STRING_ARG
 	{
 		OUTYY(("P(server_dlv_anchor_file:%s)\n", $2));
-		free(cfg_parser->cfg->dlv_anchor_file);
-		cfg_parser->cfg->dlv_anchor_file = $2;
+		log_warn("option dlv-anchor-file ignored: DLV is decommissioned");
```
ralphdolmans, line 1123: $2 can be free'd here, also in `server_dlv_anchor`

ralphdolmans (approved):
`val_stub_noroot.rpl` still sets a `dlv-anchor`. No big deal as it is not needed for the test to succeed.

No other comments, looks good to me.

wcawijngaards:
Thanks for the reviews, fixed the comments and merged it.

PR #306, opened 2020-09-10
Fix: PID file shall be owned by root

### LINKED ISSUE #303: /run/unbound.pid chowned to unbound:unbound
Under Devuan or Debian-with-sysvinit, unbound chowns its pidfile to the
daemon:

#  ifdef HAVE_CHOWN
                        if(chown(daemon->pidfile, cfg_uid, cfg_gid) == -1) {
                                verbose(VERB_QUERY, "cannot chown %u.%u %s: %s",
                                        (unsigned)cfg_uid, (unsigned)cfg_gid,
                                        daemon->pidfile, strerror(errno));
                        }
#  endif /* HAVE_CHOWN */

This is problematic, though. From start-stop-daemon's man page:

       -p, --pidfile pid-file
              Check whether a process has created the file pid-file.

              Note:  using  this  matching option alone might cause unintended
              processes to be acted on, if the old process terminated  without
              being able to remove the pid-file.

              Warning:  using  this match option with a world-writable pidfile
              or using it alone with a daemon that writes the  pidfile  as  an
              unprivileged  (non-root)  user  will  be  refused  with an error
              (since version 1.19.3) as  this  is  a  security  risk,  because
              either  any  user  can  write  to  it,  or  if  the  daemon gets
              compromised, the contents of the pidfile cannot be trusted,  and
              then  a  privileged  runner  (such as an init script executed as
              root)  would  end  up  acting  on  any  system  process.   Using
              /dev/null is exempt from these checks.

And sure enough, the shipped init script doesn't actually stop the daemon, 
and restarts don't populate the pid file but do leave multiple daemons 
running.

It's unclear to me why the pidfile would ever want to be chown'd. I see in 
the changelogs:

26 January 2016: Wouter
        - Fix #734: chown the pidfile if it resides inside the chroot.

...but I'm not sure where that buglist is to gather context. Seems like a
pidfile should explicitly not be inside a chroot under most circumstances,
unless the daemon is trying to relaunch itself chrooted...?

Either not chowning at all or having a runtime knob to control this 
chowning would be useful fixes. Thoughts?

Berbe:
Fixes #303

start-stop-daemon requires PID files to be owned by root as a security feature
Hence, required for SysVinit compatibility

### DIFF
--- a/daemon/unbound.c
+++ b/daemon/unbound.c
@@ -507,16 +507,6 @@ perform_setup(struct daemon* daemon, struct config_file* cfg, int debug_mode,
 #ifdef HAVE_KILL
 	if(cfg->pidfile && cfg->pidfile[0] && need_pidfile) {
 		writepid(daemon->pidfile, getpid());
-		if(cfg->username && cfg->username[0] && cfg_uid != (uid_t)-1 &&
-			pidinchroot) {
-#  ifdef HAVE_CHOWN
-			if(chown(daemon->pidfile, cfg_uid, cfg_gid) == -1) {
-				verbose(VERB_QUERY, "cannot chown %u.%u %s: %s",
-					(unsigned)cfg_uid, (unsigned)cfg_gid,
-					daemon->pidfile, strerror(errno));
-			}
-#  endif /* HAVE_CHOWN */
-		}
 	}
 #else
 	(void)daemon;


gthess:
Closing as the issue is already resolved.

PR #311, opened 2020-09-19
Dynlibmod leak

luismerino:
Hi! A memory leak and unchecked malloc fixed in dynlibmod. Does it look good?


wcawijngaards (approved):
Looks good, that removes memory leaks on failure cases.

PR #326, opened 2020-10-14
DoH: implement content-length header field

netblue30:
Simple implementation based on the existing code.


wcawijngaards:
Thanks for the patch!  Applied it, made small fixes to avoid warnings and portability.

PR #340, opened 2020-11-09
rpz clientip triggers and tcp-only action

magenbluten:
please review. this  is a draft.

provides:
- [ ] nsdname, nsip, clientip triggers
- [ ] tcp-only, passthru, drop actions

open:
- [x] changes to the man page(s)
- [x] fix the precedence of applied triggers such that they correspond to the rfc
- [x] correct the design decision of what domain name to match at a given delegation point (nsdname)


services/rpz.c
```diff
@@ -938,107 +1293,692 @@ log_rpz_apply(uint8_t* dname, enum rpz_action a, struct query_info* qinfo,
...
+					     LDNS_RCODE_NOERROR | BIT_RD | BIT_QR | BIT_AA | BIT_RA,
+					     1, //qd
+					     0, //ttl
+					     0, //prettl
+					     0, //expttl
+					     0, //an
+					     0, //ns
+					     0, //ar
+					     0, //total
+					     sec_status_secure);
```
wcawijngaards, line 1515: Use sec_status_insecure, and also at 1535 and 1563, just like respip.c has sec_status_insecure in 664 (make_new_reply_info).  The message is not one with dnssec, so we can use this sec status.

magenbluten: thank you for your review. another question would be if the `reply->is_authoritative` flag should be set as well?

wcawijngaards: Actually that would a great idea.  That flag did not exist in earlier versions where people wanted to 'insert an item in cache', but it does make a lot of sense and makes for better output.  I also left review comments on the unbound chat channel that you should look at w.r.t. to some connection questions where code goes.

services/rpz.c
```diff
@@ -938,107 +1293,692 @@ log_rpz_apply(uint8_t* dname, enum rpz_action a, struct query_info* qinfo,
...
+	sldns_buffer* buf, struct regional* temp,
+	/* output parameters */
+	struct local_zone** z_out, struct auth_zone** a_out, struct rpz** r_out)
+{
+	int ret = 0;
+	enum rpz_action client_action;
+	struct clientip_synthesized_rr* node = rpz_resolve_client_action_and_zone(
+		az, qinfo, repinfo, taglist, taglen, stats, z_out, a_out, r_out);
+
+	client_action = node == NULL ? RPZ_INVALID_ACTION : node->action;
```
wcawijngaards, line 1912: I would suggest parentheses around the if.  `(node == ... action);`  I worry that `=` is assigned `(node)` or `(node==NULL)`.

wcawijngaards:
Hi Alex, I do not have any further comments at this time, it is nice code!
Also layout looks nice, even comments in function definitions are good looking, not sure which method you ended up with.

services/rpz.c
wcawijngaards: Was fixed, closing it.

services/rpz.c
wcawijngaards: This is fixed in commit https://github.com/NLnetLabs/unbound/commit/9f40e93fe6f1473cd2bec93416b397599ce03fc2 .

wcawijngaards:
The code from this pull request moved into the #401 pull request, that was merged and this also closed and merged this pull request.

PR #341, opened 2020-11-09
fixing a possible memory leak.

ihsinme:
the form of using the realloc function can lead to the loss of a pointer to previously allocated memory. 

the suggested fixes will eliminate this issue.


wcawijngaards:
Thanks!  Applied your fix, it only runs once at the start of the program so that would not really impact operations, but it is good to fix the error case.

PR #351, opened 2020-11-21
Add AF_NETLINK to set of allowed socket address families

### LINKED ISSUE #350: 1.12.0 error: failed to list interfaces: getifaddrs: Address family not supported by protocol
Hi! I'm packaging unbound for Arch Linux.
I am currently trying to upgrade the package to 1.12.0. Unfortunately I am running into serious issues with the service right on start.
With my configuration file, compatible with 1.11.0 I am trying to start unbound.service, but it returns:

```
unbound[1610946:0] error: failed to list interfaces: getifaddrs: Address family not supported by protocol
unbound[1610946:0] fatal error: could not open ports
```

I see that f6a527c25ad2e60e2dc129fff3605e6ec48c30f2 has changed the behavior of the `*interface:` settings in the configuration. Unfortunately, this doesn't seem to work out of the box with the systemd unit anymore (or because of another problem?).

Even if I provide interface names to the respective `*interface:` configuration settings, I get the above error message.

dvzrv:
contrib/unbound{,_portable}.service.in:
With the changes introduced in f6a527c25ad2e60e2dc129fff3605e6ec48c30f2
it is now necessary to also allow access to the AF_NETLINK socket
address family to be able to get information from interfaces.

Without the AF_NETLINK address family the systemd service errors with:

```
error: failed to list interfaces: getifaddrs: Address family not
supported by protocol
```

Fixes #350


wcawijngaards:
Thanks for the fix, moved it into the code base!

PR #363, opened 2020-12-02
Support for RFC5001: DNS Name Server Identifier (NSID) Option

wtoorop:
Support for RFC5001: DNS Name Server Identifier (NSID) Option with a nsid: option in unbound.conf


wcawijngaards (approved):
Looks good!  Nothing in example.conf.in for new option.

gthess (approved):
Same here!
Although some tests a la https://github.com/NLnetLabs/unbound/blob/master/testdata/edns_client_string.rpl would be useful.

PR #367, opened 2020-12-09
DNSTAP log local address

### LINKED ISSUE #365: DNSTAP logging in Unbound-1.11.0
Hello, Wouter!

As i said earlier i have made patches set for Unbound-1.11.0 to log over DNSTAP both addresses - source (from client) address and destination (local or service) one. My core idea is to introduce NSD-like structure called unbound_socket and use it in comm_points of netevent code. So please look in my patches, i hope they will be reasonable and secure for application.
And only one TODO that i have to do in future - right now i can log only the front side, i.e. client requests and responses. Upstream side is currently logged as 0.0.0.0 local address. Its a question of some rework of outside_network code.

All the patches are in attached tarball.

Thank you in advance!
[patches.tar.gz](https://github.com/NLnetLabs/unbound/files/5659923/patches.tar.gz)

### LINKED ISSUE #368: dnstap does not log the DNS message ID for FORWARDER_QUERY
I'm seeing an ID of 0 for FORWARDER_QUERY dnstap messages. The FORWARDER_RESPONSE does contain the ID as do all of the CLIENT_QUERY and RESOLVER_QUERY message types. See example below (decoded using https://github.com/dnstap/golang-dnstap/tree/master/dnstap). Shouldn't the forwarder query have a non-zero ID? I'm using unbound 1.13.0.

```
type: MESSAGE
message:
  type: FORWARDER_QUERY
  query_time: !!timestamp 2020-12-09 21:40:22.415342
  socket_family: INET
  socket_protocol: TCP
  response_address: 1.1.1.1
  response_port: 853
  query_zone: "."
  query_message: |
    ;; opcode: QUERY, status: NOERROR, id: 0   <------------------------------------ ID is 0?
    ;; flags: rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

    ;; QUESTION SECTION:
    ;itunes.apple.com.edgekey.net.      IN       HTTPS

    ;; ADDITIONAL SECTION:

    ;; OPT PSEUDOSECTION:
    ; EDNS: version 0; flags: do; udp: 1472
```

```
type: MESSAGE
message:
  type: FORWARDER_RESPONSE
  query_time: !!timestamp 2020-12-09 21:40:22.414892
  response_time: !!timestamp 2020-12-09 21:40:22.44134
  socket_family: INET
  socket_protocol: TCP
  response_address: 1.1.1.1
  response_port: 853
  query_zone: "."
  response_message: |
    ;; opcode: QUERY, status: NOERROR, id: 6959   <------------------------------------ ID is non-zero
    ;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 1, ADDITIONAL: 1

    ;; QUESTION SECTION:
    ;itunes.apple.com.edgekey.net.      IN       HTTPS

    ;; ANSWER SECTION:
    itunes.apple.com.edgekey.net.       21423   IN      CNAME   e673.dsce9.akamaiedge.net.

    ;; AUTHORITY SECTION:
    dsce9.akamaiedge.net.       823     IN      SOA     n0dsce9.akamaiedge.net. hostmaster.akamai.com. 1607549848 1000 1000 1000 1800

    ;; ADDITIONAL SECTION:

    ;; OPT PSEUDOSECTION:
    ; EDNS: version 0; flags: do; udp: 1232
    ; PADDING: 0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

wcawijngaards:
This PR contains the code contributed in issue #365 that is imported and then gets modified for insertion in to the code repository.

The code adds logging of the destination, or local, address to the dnstap logging.


services/outside_network.c
```diff
@@ -1914,11 +1940,20 @@ randomize_and_send_udp(struct pending* pend, sldns_buffer* packet, int timeout)
 	comm_timer_set(pend->timer, &tv);
 
 #ifdef USE_DNSTAP
+	/*
+	 * sending src (local service)/dst (upstream) addresses over DNSTAP
+	 * TODO: right now there are no chances to get the src (local service) addr. So we will pass 0.0.0.0 (::)
+	 * to argument for dt_msg_send_outside_query()/dt_msg_send_outside_response() calls.
+	 * For the both UDP and TCP.
```
gthess: It seems this comment is now outdated.

services/listen_dnsport.c
```diff
@@ -916,7 +926,7 @@ static int
 make_sock(int stype, const char* ifname, const char* port, 
 	struct addrinfo *hints, int v6only, int* noip6, size_t rcv, size_t snd,
 	int* reuseport, int transparent, int tcp_mss, int nodelay, int freebind,
-	int use_systemd, int dscp)
+	int use_systemd, int dscp, struct unbound_socket** ub_sock)
```
gthess: We could do with a simple pointer in this call, right?

services/listen_dnsport.c
```diff
@@ -967,7 +981,7 @@ static int
 make_sock_port(int stype, const char* ifname, const char* port, 
 	struct addrinfo *hints, int v6only, int* noip6, size_t rcv, size_t snd,
 	int* reuseport, int transparent, int tcp_mss, int nodelay, int freebind,
-	int use_systemd, int dscp)
+	int use_systemd, int dscp, struct unbound_socket** ub_sock)
```
gthess: Same here for a simple pointer.

gthess (approved):
Looks good to me! Nits only.

services/outside_network.c
wcawijngaards: Yes, but also no, in some cases the user could have configured IP-addresses, but when unbound starts with defaults, or with config of the wildcard, the 0.0.0.0 is in the addr pointer that is passed here, and it is not the address of the interface that the OS chooses after unbound uses that wildcard.

services/listen_dnsport.c
wcawijngaards: Fixed in https://github.com/NLnetLabs/unbound/pull/367/commits/6cda81b7e487afa35fe723eb0f92c6ce6ea42f7c  Thanks for the improvement, that is neater.

services/listen_dnsport.c
wcawijngaards: Fixed in https://github.com/NLnetLabs/unbound/pull/367/commits/6cda81b7e487afa35fe723eb0f92c6ce6ea42f7c  with the neater code.

services/outside_network.c
gthess: Then maybe:
```
There are no chances to get the src (local service) addr if unbound is not configured with specific outgoing IP-addresses. So we will pass 0.0.0.0 (::) to argument for dt_msg_send_outside_query()/dt_msg_send_outside_response() calls.
```
And also remove the "For the both UDP and TCP." line since here we are only doing UDP?

wcawijngaards: Fixed it like you suggested in https://github.com/NLnetLabs/unbound/pull/367/commits/04b4f515a7b3de1dc02d78990f7a0c3d1d6d84aa    I think that is a better comment.  It is also true for TCP, but we need not spam it everywhere we call this function.

iruzanov:
Hello, Wouter!

In the previous week  i've downloaded tarball with source code of unbound-1.13.1 and did not find the code concerning of DNSTAP logging of local IP. Has this code removed from this version of Unbound? Why i'm asking - i'm just worried about future upgrades to the recent versions of Unbound. And in our project the logging of local IP over DNSTAP is of great importance.

Big thank you in advance for answering!

wcawijngaards:
The code is present in the code repository, but has not been in a release tarball yet.  The 1.13.1 release was from before the code merge.

iruzanov:
I got it!
Thank you very much for the clarification, Wouter!

PR #391, opened 2021-01-01
Add start_time to reply callbacks so modules can compute the response…

fhriley:
… time.


wcawijngaards (approved):
The struct timeval should be passed with a ptr instead of a structure function argument.  This is much easier for function signature, also for dynamic libraries and different compilers.  The python start time calculation has a casting issue, losing fractions.  Indentation (but that may be because of the view here).  I approve of the changes, allows modules to access more information.  The callbacks get either 0 or the value from the packet if also packets exist for that query, but that seems fine now, not really a need to add a timer value to the callbacks.

wcawijngaards:
Thanks for the patch!  Applied with a number of small changes.

wcawijngaards:
As an aside, the cast in the python start time looks like it was fine upon a further look; but I added explicit casts.

Issue #376, opened 2020-12-14
SIGSEGV on FreeBSD with version 1.13.0


kapfenho:
With version 1.13.0 I got a segmentation violation (sorry, no core file). That's the first time, after years of stability.

Last messages:
```
Dec 14 01:03:59 <daemon.notice> homer unbound[72482]: [72482:0] notice: sendto failed: Permission denied  
Dec 14 01:03:59 <kern.info> homer kernel: pid 72482 (unbound), jid 0, uid 59: exited on signal 11  
```

System:
```
FreeBSD homer 12.2-RELEASE-p1 FreeBSD 12.2-RELEASE-p1 GENERIC  amd64
```

Unbound config (using IPv4 & IPv6):
```
server:
        verbosity: 1
        statistics-interval: 3600
        interface: ::1
        interface: ...
        interface-automatic: yes
        outgoing-interface: ...
        access-control: ... allow
        root-hints: "/usr/local/etc/unbound/root.hints"
        hide-identity: yes
        hide-version: yes
        harden-algo-downgrade: yes
        auto-trust-anchor-file: "/usr/local/etc/unbound/root.key"
remote-control:
        control-enable: yes
        control-interface: 127.0.0.1
```

rozhuk-im:
Crash for me too, build from sources: TFOCL, TFOSE, THREADS options is set.

kapfenho:
Options here: ECDSA GOST LIBEVENT THREADS

emikulic:
Some debugging and backtraces in https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=251821

delphij:
I think this was caused by commit 15e8f5c6d4f85c3858c7ad8927ee50f2da8a8a54 .

The problem is that the logging code really expects a valid `addr` pointer, and now in `udp_connect` case it's no longer true.

I think this would stop unbound from crashing for the FreeBSD case:

```
diff --git a/services/outside_network.c b/services/outside_network.c
index 0886907f..e2843573 100644
--- a/services/outside_network.c
+++ b/services/outside_network.c
@@ -1901,6 +1901,9 @@ randomize_and_send_udp(struct pending* pend, sldns_buffer* packet, int timeout)
 	/* send it over the commlink */
 	if(outnet->udp_connect) {
 		if(!comm_point_send_udp_msg(pend->pc->cp, packet, NULL, 0)) {
+			log_addr(VERB_OPS, "remote address is",
[...]

PR #395, opened 2021-01-06
add missing null check

mptre:
I have a unbound forward zone configured on my router for my $DAYJOB.
The address associated with the zone is only accessible when the router
is connected to a VPN. If the VPN connection is absent, trying to
resolve any domain that must be handled by the zone crashes unbound.
Turns out there's a missing NULL check in `comm_point_send_udp_msg()`.
The same routine already has `if (addr) {} else {}` branches so I guess
protecting the call to `log_addr()` using the same conditional is
reasonable

I have also committed the same fix to unbound shipped with OpenBSD[1].

[1] https://marc.info/?l=openbsd-cvs&m=160993335615698&w=2

### DIFF
--- a/util/netevent.c
+++ b/util/netevent.c
@@ -388,8 +388,9 @@ comm_point_send_udp_msg(struct comm_point *c, sldns_buffer* packet,
 		} else {
 			verbose(VERB_OPS, "send failed: %s", sock_strerror(errno));
 		}
-		log_addr(VERB_OPS, "remote address is",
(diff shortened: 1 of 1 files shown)


wcawijngaards:
Thank you for the patch.  I think the same issue was already solved in a fix for #376 in commit https://github.com/NLnetLabs/unbound/commit/08968baec1122a58bb90d8f97ad948a75f8a5d69 .  But the patch looks fine, so merged it.

Issue #394, opened 2021-01-05
Prefetch logic is called when serve expired is turned on but blacklists cache


edevil:
Commit https://github.com/NLnetLabs/unbound/commit/a8db52120b3d3ac078601ed875e8eac92bbbae65 made the prefetch logic be used when a cache entry with an expired TTL is served. However, prefetch blacklists the cache which makes subsequent requests for that record not able to use the cache, which defeats the purpose of `serve-expired`.

gthess:
It's not clear to me what the issue is.
serve_expired needs to always try and update the cache, otherwise the expired records will never be updated.

That commit fixed the comparison i.e., when serve_expired is used compare the TTL, when prefetch is used compare the calculated prefetch_ttl.

You are right that the prefetch logic blacklists the cache because it tries to fetch new data, but it is only blacklisted for that query in flight (and queries that may indirectly rely on this query). Other clients *reaching* unbound don't have the cache blacklisted.

Do you have an example and/or configuration that you encounter what you describe?

edevil:
That is not the behavior I observe. I tried with 1.13.0 and (e55f38fa83f5d43f202ca489cc72078adeeb694f) compiled on MacOS and this simple config: 
```
server:
    verbosity: 10
    port: 5354
    do-daemonize: no
    log-time-ascii: yes
    log-queries: yes
    log-replies: yes
    log-tag-queryreply: yes
    log-local-actions: yes
    log-servfail: no
    username: ""
    logfile: /Users/acruz/work/unbound-work/unbound.log
    use-syslog: no
    chroot: ""
    serve-expired: yes
    serve-expired-ttl: 3600
    serve-expired-client-timeout: 500
    directory: /Users/acruz/work/unbound-work
    access-control: 127.0.0.0/8 allow_snoop


remote-control:
    control-enable: yes
    control-interface: /Users/acruz/work/unbound-work/control.socket
```
What I see is that when an entry expires from the cache, the next client will get served the stale entry immediately due to `reply_and_prefetch()`. But if upstream is slow or unreachable subsequent unbound clients find the cache blacklisted and are kept waiting for prefetch to finish, if it ever finishes.

gthess:
I believe I now see what is happening.

You have serve-expired-client-timeout enabled. This means that serving expired data is a back up option when resolving has taken more than 500 ms in your case.
[...]

PR #396, opened 2021-01-06
Revert "- Fix the relationship between serve-expired and prefetch opt…

edevil:
…ions,"

This reverts commit a8db52120b3d3ac078601ed875e8eac92bbbae65.

I am not quite sure of the rationale for the previous patch, but with it applied what I see is that when an expired entry is fetched from the cache the prefetch routine `reply_and_prefetch` is called. When this happens a response is returned to the client, however, the current cache entry is blacklisted. This makes the cache unusable until a response is fetched from the authoritative server.

Since the authoritative server can be down or slow, we would want to serve the stale response while this happens, and with this commit reverted that's exactly what I see.

Again, I don't know what this commit attempted to solve but the serve expired feature seems to work better without it.

Addresses #394

### DIFF
--- a/daemon/worker.c
+++ b/daemon/worker.c
@@ -1487,11 +1487,9 @@ lookup_cache:
 				 * Note that if there is more than one pass
 				 * its qname must be that used for cache
 				 * lookup. */
-				if((worker->env.cfg->prefetch && *worker->env.now >=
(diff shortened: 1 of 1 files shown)


gthess:
Thanks for looking into this but as described in #394, I don't think this is the right approach.
Closing.

PR #399, opened 2021-01-13
The lock of  lruhash table should unlocked after markdel entry

rma-shuyu:
…hen lru_remove the entry , then unlock the table, and then markdel entry , but in function rrset_cache_touch , the entry will be touched to lru again before markdelling entry in function lruhash_remove. This is a bug!

### DIFF
--- a/util/storage/lruhash.c
+++ b/util/storage/lruhash.c
@@ -398,13 +398,13 @@ lruhash_remove(struct lruhash* table, hashvalue_type hash, void* key)
 		return;
 	}
 	table->num--;
-	table->space_used -= (*table->sizefunc)(entry->key, entry->data);
-	lock_quick_unlock(&table->lock);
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
Changes look good.

wcawijngaards:
Previous title was, from xiangbao227: ```I found that in function "lruhash_remove", table was locked at first ,then lru_remove the entry , then unlock the table, and then markdel entry , but in function "rrset_cache_touch" , the entry will be touched to lru again before markdelling entry in function "lruhash_remove". This may lead to core!```
Changed the title to a shorter one.

wcawijngaards:
Thank you for the bug fix!  That was hard to find, I think.  I have committed the fix to the code repository.

The other calls of the markdel function have the table lock held, so they should not have this race condition.

PR #407, opened 2021-01-19
Fix TTL of SOA record for negative answers

gthess:
This affects localzone and authzone data. The TTL is now the minimum of the SOA TTL and the SOA.MINIMUM.

### DIFF
--- a/services/localzone.c
+++ b/services/localzone.c
@@ -463,6 +463,48 @@ lz_find_create_node(struct local_zone* z, uint8_t* nm, size_t nmlen,
 	return 1;
 }
 
+/* Mark the SOA record for the zone. This only marks the SOA rrset; the data
+ * for the RR is entered later on local_zone_enter_rr() as with the other
+ * records. An artifical soa_negative record with a modified TTL (minimum of
+ * the TTL and the SOA.MINIMUM) is also created and marked for usage with
+ * negative answers and to avoid allocations during those answers. */
+static int
+lz_mark_soa_for_zone(struct local_zone* z, struct ub_packed_rrset_key* soa_rrset,
+	uint8_t* rdata, size_t rdata_len, time_t ttl, const char* rrstr)
+{
+	struct packed_rrset_data* pd = (struct packed_rrset_data*)
+		regional_alloc_zero(z->region, sizeof(*pd));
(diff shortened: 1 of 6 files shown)


testdata/auth_zonefile_down.rpl
```diff
@@ -1,6 +1,12 @@
 ; config options
 server:
 	target-fetch-policy: "0 0 0 0 0"
+	; Options for singed zone. The zone is partially copied from val_negcache_nxdomain.rpl
```
wcawijngaards:
```suggestion
	; Options for signed zone. The zone is partially copied from val_negcache_nxdomain.rpl
```

services/localzone.c
```diff
@@ -463,6 +463,48 @@ lz_find_create_node(struct local_zone* z, uint8_t* nm, size_t nmlen,
...
+	rrset_negative->rk.rrset_class = soa_rrset->rk.rrset_class;
+	if(!rrset_insert_rr(z->region, pd, rdata, rdata_len, ttl, rrstr))
+		return 0;
+	/* last 4 bytes are minimum ttl in network format */
+	if(pd->count == 0 || pd->rr_len[0] < 2+4) {
+		return 0;
+	}
+	minimum = (time_t)sldns_read_uint32(pd->rr_data[0]+(pd->rr_len[0]-4));
+	pd->ttl = ttl<minimum?ttl:minimum;
+	pd->rr_ttl[0] = pd->ttl;
```
wcawijngaards: You are missing out on the rrsig ttl's here, that are fixed in the authzone code.

wcawijngaards (approved):
Nice stuff, approve of it!

testdata/auth_zonefile_down.rpl
gthess: Fixed

services/localzone.c
gthess: I am not sure if rrsigs are put under the relevant rrsets for local data. From what I see in the code and what I tried I can't get an rrsig answer from local data, unless I ask for the rrsig type.

gthess:
Thanks Wouter! One comment is still unresolved but I don't think it can be resolved unless I am missing something. After that I am ready to merge.

services/localzone.c
wcawijngaards: Yes you must be right, that solves it.

Issue #257, opened 2020-06-29
unbound-anchor default configuration file


sibeream:
Hello everyone,

There is a possible conflict for unbound-anchor used ports with SELinux policy.
There is a bug filed in Red Hat Bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1830625

In that particular example, unbound-anchor tries to use port 61000, which is not allowed by the SELinux policy.
I was looking for a way to provide a default configuration file to avoid that port, but only found a way to provide configuration using -C option.

I would like to propose the following behavior:
1. If a configuration file is specified with -C option, use specified one
2. else If /etc/unbound/unbound-anchor.conf file exists, use it
3. else don't use a configuration file

I believe this approach will make the tool more flexible in terms of different Linux distributions since they can have different SELinux policies, so maintainers will be able to create proper configurations.
I can prepare a patch to implement that behavior. I just want to be sure that this approach makes sense for the community.

gthess:
Hi,
Since the option to use a configuration file with `-C` is already there I don't see why we would need a specific configuration file for unbound-anchor.

pemensik:
Hi @gthess , we consider duplication of values already set in the system as inefficient. We would have to include some ranges in a configuration file. We currently do not use configuration file for _unbound-anchor_ or _unbound-host_. They all should use the range.

Because that range could be modified by a system administrator, it could be different on some machines. Administrator would have to change sysctl ip_local_port_range AND to adjust matching values both in unbound.conf, unbound-anchor.conf and maybe some others. All such changes would have to be done manually, instead of just following the system settings.

Would _configure_ option to enable such behaviour work better? It might not be enabled by default, but we would like to enable it instead of providing multiple configuration files with fixed numbers.

pemensik:
I have to add, ~30k possible values should be more than enough for tools like unbound-anchor or unbound-host. We would like to still allow overriding autodetected ranges from configuration file, but without mandatory settings on selinux enabled machines.

This change should also affect any third party tools linking to unbound libraries, such as libreswan, getdns and openvswitch. I think it is better to have good defaults than to force manual configuration to all those tools.

gthess:
Hi @pemensik, first of all I appreciate the further feedback.

It is my understanding that this kernel range is used by the kernel when a program does not provide its own port. In Unbound's case, for UDP communication, Unbound picks its own ports as port randomization is an important part of DNS over UDP.
It is my understanding that this range is also used by SELinux for convenience.

[...]

PR #409, opened 2021-01-20
Use /proc/sys/net/ipv4/ip_local_port_range to determine available outgoing ports

sibeream:
Use /proc/sys/net/ipv4/ip_local_port_range file to determine range of available ports.
Relates to #257

### DIFF
--- a/util/config_file.c
+++ b/util/config_file.c
@@ -1568,6 +1568,31 @@ init_outgoing_availports(int* a, int num)
 		if(iana_assigned[i] < num)
 			a[iana_assigned[i]] = 0;
 	}
+
+#ifndef UB_ON_WINDOWS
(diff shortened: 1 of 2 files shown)


gthess:
Thanks for taking the time to tackle this but I am afraid I have to refuse it.

From the [kernel documentation](https://man7.org/linux/man-pages/man7/ip.7.html) this range is meant to be used when the kernel picks an outgoing port. In the case of unbound, port selection is done by unbound it self. Also by using a far broader range (ip_local_port_range in my system and the one I see configured in SELinux is only ~30K) there are more random bits for outgoing UDP ports which is crucial for DNS.

I understand that this could pose a problem when SELinux is configured but on those systems the already available configuration options `outgoing-port-permit` and `outgoing-port-avoid`could be used to influence port selection.

PR #420, opened 2021-02-10
DOH not responsing with "http2_query_read_done failure" logged

dyunwei:
I found that Unbound will drop normal queries when DOH is enabled. After some investigation i have found the root cause.

In function mesh_query_done, assign r->query_reply.c->buffer to r_buffer, and then call mesh_send_reply. The r_buffer is used to fill the dns reply in the mesh_send_reply function. then call comm_point_send_reply, and then call http2_submit_dns_response to send the DOH response. However, the buffer is not cleared after use. If the query length is greater than the last response length, the next dns query in the same H2 session will be dropped, and then the debug message "http2_query_read_done failure: can't fit qbuffer in c->buffer" be logged.

Clear the buffer after use can fix the issue.


wcawijngaards:
Thanks! That looks good, merged it into the code repository.

PR #427, opened 2021-02-16
Fix #426: Replace _Py_fopen() with fopen() in pythonmod.c

### LINKED ISSUE #426: unbound uses private _Py_fopen() function, removed in Python 3.10
unbound uses the private `_Py_fopen()` function, but this function was removed in Python 3.10. I wrote a PR to replace it with `fopen()` which is always available.

With issue #378 fix, it should make unbound compatible with Python 3.10.

vstinner:
The private _Py_fopen() function has been removed in Python 3.10.

Fix build on Python 3.10.

### DIFF
--- a/doc/Changelog
+++ b/doc/Changelog
@@ -2,6 +2,7 @@
 	- Fix #422: IPv6 fallback issues when IPv6 is not properly
 	  enabled/configured.
 	- Fix to make tests work with support indicators set for iterator.
+	- Fix build on Python 3.10.
 
 10 February 2021: Wouter
 	- Merge PR #420 from dyunwei: DOH not responsing with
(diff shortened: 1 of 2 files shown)


vstinner:
@wcawijngaards: Hi, would you mind to review this simple fix? See #426 for the rationale. With this change, unbound should build on Python 3.10.

vstinner:
FYI I created https://src.fedoraproject.org/rpms/unbound/pull-request/6 to test this change on Fedora Rawhide with Python 3.10, to check if the unbound package builds successfully on Python 3.10.

wcawijngaards:
Looks good!  Thank you for the bug fix.

vstinner:
You're welcome ;-)

PR #440, opened 2021-03-09
Various fixes to contrib/unbound_munin_ file

kimheino:
The most important fix is "fix statistics after unbound restart / server reboot". Other changes are mostly cosmetic.

### DIFF
--- a/contrib/unbound_munin_
+++ b/contrib/unbound_munin_
@@ -14,7 +14,6 @@
 # Run the command unbound-control-setup to generate the key files.
 #
 # Environment variables for this script
-#	statefile	- where to put temporary statefile.
 #	unbound_conf	- where the unbound.conf file is located.
 #	unbound_control	- where to find unbound-control executable.
 #	spoof_warn	- what level to warn about spoofing
@@ -24,7 +23,6 @@
 # with:
 # [unbound*]
 # user root
-# env.statefile /usr/local/var/munin/plugin-state/unbound-state
 # env.unbound_conf /usr/local/etc/unbound/unbound.conf
 # env.unbound_control /usr/local/sbin/unbound-control
 # env.spoof_warn 1000
@@ -66,7 +64,6 @@ System with unbound daemon.
 
   [unbound*]
   user root
-  env.statefile /usr/local/var/munin/plugin-state/unbound-state
   env.unbound_conf /usr/local/etc/unbound/unbound.conf
   env.unbound_control /usr/local/sbin/unbound-control
   env.spoof_warn 1000
@@ -98,7 +95,8 @@ BSD
 
 =cut
 
-state=${statefile:-/usr/local/var/munin/plugin-state/unbound-state}
+. ${MUNIN_LIBDIR}/plugins/plugin.sh
+state="${MUNIN_PLUGSTATE}/unbound.state"
 conf=${unbound_conf:-/usr/local/etc/unbound/unbound.conf}
 ctrl=${unbound_control:-/usr/local/sbin/unbound-control}
 warn=${spoof_warn:-1000}
--- a/contrib/unbound_munin_
+++ b/contrib/unbound_munin_
@@ -126,7 +126,7 @@ get_state ( ) {
 	# obtain lock for fetching the state
 	# because there is a race condition in fetching and writing to file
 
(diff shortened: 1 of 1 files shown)


contrib/unbound_munin_
```diff
@@ -98,7 +95,9 @@ BSD
 
 =cut
 
-state=${statefile:-/usr/local/var/munin/plugin-state/unbound-state}
+. ${MUNIN_LIBDIR}/plugins/plugin.sh
```
gthess: Why do we need plugin.sh? I don't think we use anything from there, right?

contrib/unbound_munin_
```diff
@@ -98,7 +95,9 @@ BSD
...
+state="${MUNIN_PLUGSTATE}/unbound.state"
+seentags="${MUNIN_PLUGSTATE}/unbound-seentags.state"
```
gthess: Are these environment variables guaranteed to be set?

contrib/unbound_munin_
```diff
@@ -467,27 +475,23 @@ memory)
...
+	for nm in `grep "^num.query.opcode" $seentags`; do
+		print_value $nm
 	done
 	;;
 by_rcode)
-	for x in `grep "^num.answer.rcode" $state`; do
-		nm=`echo $x | sed -e 's/=.*$//'`
-		print_value_line $nm $x
+	for nm in `grep "^num.answer.rcode" $seentags`; do
+		print_value $nm $x
```
gthess: stray $x?

contrib/unbound_munin_
```diff
@@ -121,12 +120,24 @@ get_value ( ) {
...
+update_seentags() {
+    tmplist="$(cat ${seentags} 2> /dev/null)
+num.query.type.A
+num.query.class.IN
+num.query.opcode.QUERY
+num.answer.rcode.NOERROR
+"
+    (echo "${tmplist}"; grep ^num ${state} | sed -e 's/=.*//') | sort -u > ${seentags}
+}
+
```
gthess, line 133: OK, so here is the core of the change. If I understand correctly you want munin to keep plotting tags that unbound was reporting on before the restart (with 0 as the value if the tag is no longer reported by unbound; based on get_value()).

But why do you always set the above 4?

gthess (commented):
Thanks for taking the time for this!
I have some questions inline before accepting this.

Not a munin user but I am worried if this change catches other users by surprise or if it is not relevant to other users.
Do you see value in having an environment variable to toggle this new behavior? Then based on the environment variable update_seentags() could just write the state content (filtered) to the seentags file and the old behavior is also kept by default.

kimheino:
> Not a munin user but I am worried if this change catches other users by surprise or if it is not relevant to other users.
> Do you see value in having an environment variable to toggle this new behavior? Then based on the environment variable update_seentags() could just write the state content (filtered) to the seentags file and the old behavior is also kept by default.

Old behavior has two fatal errors:

1) Munin reports error if you restart unbound (tag vanishes). 
2) New tags will not be included to Munin's graph if they are not present at the first run of munin-async. So basically graph stays empty if you reboot and are using munin-async.

New behavior might include some tags that stays zero forever, but that should not matter. So no, I don't see any point including config option for switching between old/new.

I'll reply inline for others.

contrib/unbound_munin_
kimheino: You're correct, I'll remove it.

contrib/unbound_munin_
kimheino: Yes, they are always present.

contrib/unbound_munin_
kimheino: Oops, sorry, I'll remove it.

contrib/unbound_munin_
kimheino: Yes, you understood it correctly.

Those 4 tags are needed, as later on code has "`for nm in grep "^num.query.type" $seentags; do`" (and num.query,class, opcode and rcode). This makes it sure that there are at least one tag for each graph, needed by Munin. Graph without tag is not allowed.

gthess:
LGTM now, thanks for explaining the reason behind the changes.

PR #449, opened 2021-03-21
build: Add missing linker flags.

orbea:
When building unbound with `--with-pyunbound` and slibtool (https://dev.midipix.org/cross/slibtool) it fails with many python undefined references.

This is because the `$(LIBTOOL)` linker command has `-no-undefined` and forgot `$(LIB)` which when added solves the problem. GNU libtool does not expose this issue because it silently ignores `-no-undefined`.

Full build log: [unbound.slibtool.log](https://github.com/NLnetLabs/unbound/files/6178609/unbound.slibtool.log)

Edit: Also see this downstream issue.

https://bugs.gentoo.org/777447


wcawijngaards:
Thanks for the fix! Merged it into the code repository.

PR #458, opened 2021-03-30
X86views First commit

saksham-m:
First commit for BIND style views. Changes include:
1) Configuration parser changes including cleaner config using rules (configparser.y)
2) module_env containing current_view_env. The idea is to swap out this pointer when a view is matched.


wcawijngaards:
You should put that point in the module_qstate* qstate member. This then makes it possible to have queries that are part of a particular view. The list of views, should be in the module_env, and that pointer a pointer into the list of views.

The module_env is shared with all other queries, so that would weirdly impact other queries because one of them has a view connected to it.

saksham-m:
Thanks for the feedback. I'll refactor and push again.

PR #460, opened 2021-03-31
build: Link with the libtool archive.

orbea:
The canonical way to link internal dependencies with libtool (Or slibtool) is to use the libtool archive (`.la`) file and using `-Lfoo -lbar` should be reserved for external dependencies. Note this did not actually break the build with slibtool because the build has both `-L.` and `-L.libs`, but its still better to use `libunbound.la` instead.


wcawijngaards:
Thanks! Applied the merge so that it is in the code repository.

PR #461, opened 2021-04-06
Add max-query-restarts option

### LINKED ISSUE #438: Resolving records through more than 8 CNAME fails due to hardcoded MAX_RESTART_COUNT
I've run into an edge case where a name passed through 9 CNAMEs (and 10 total resolutions to get the final answer). Because `MAX_RESTART_COUNT` is [hardcoded ](https://github.com/NLnetLabs/unbound/blob/6cd77933a3f113ea2bef7e4943f6dda6a26a39cb/iterator/iterator.h#L64) there's no way to tune the configuration to work around this situation.

The original bug report is in the pfSense bug tracker [here](https://redmine.pfsense.org/issues/11595) where detailed traces from `drill` are available.

cgallred:
Resolves #438 

This is a proposed new option to allow a user to override the hardcoded MAX_RESTART_COUNT limit.

I looked at other PRs that added config options, please let me know if I'm missing anything. I've verified that with a test config file `test.conf` containing `max-query-restarts: 9`, I'm able to successfully resolve logincdn.msauth.net with `unbound-host -C test.conf`.


steve-boyle:
I keep running into the bug in #438 on the latest version of pfSense (CE 2.6.0 as of this writing). This pull request would let me resolve the issue myself.  My only choice now is to use a different DNS resolver - other resolvers do not have this issue.  I'm happy to replace unbound but I prefer to just tweak it to work with Microsoft logins.

Unbound users cannot control what Microsoft and CDNs do with CNAMEs.  You've left us out in the cold with no solution.  You need to allow this or create a real solution for #438. As it stands I cannot use Unbound with Microsoft services.

johnnyxwan:
I have been running into this MAX_RESTART_COUNT issue for long. Silly enough, I have to create forward-zones for domains affected by this issue and forward the queries to external DNS resolver. This new option is very important and essential in my opinion, since I encounter this issue more and more frequently. Although not a scientific research, it seems a lot of CDNs are using longer and longer cname chain.

In the related issue #438, contributor of this project showed their reluctant to this change in Aug 2021. However, I would encourage the team to review this decision and hopefully merge this pull soon.

gthess:
In the end we merged this. We have been seeing more demand for a configurable option both here and from other sources.
Thanks for this :)

cgallred:
Just catching up from vacation. Thanks for accepting the change!

PR #466, opened 2021-04-19
Support OpenSSLs that lack SSL_get0_alpn_selected.

FGasper:
Unbound currently doesn’t build on CloudLinux 6. This patch seems to fix the issue, though I’m still confirming.

### DIFF
--- a/util/netevent.c
+++ b/util/netevent.c
@@ -1314,6 +1314,7 @@ ssl_handshake(struct comm_point* c)
 			c->repinfo.addrlen);
 	}
 
+#ifdef HAVE_SSL_GET0_ALPN_SELECTED
 	/* check if http2 use is negotiated */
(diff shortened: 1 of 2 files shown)


util/netevent.c
```diff
@@ -1314,6 +1314,7 @@ ssl_handshake(struct comm_point* c)
 			c->repinfo.addrlen);
 	}
 
+#ifdef HAVE_SSL_GET0_ALPN_SELECTED
```
atoomic, line 1317: nice catch!

wcawijngaards:
Thank you for the patch. Added autoconf and test to make dohclient compile, to make 'make test' succeed. That makes it compile.

If you try to use http2 unbound fails with an error that ALPN is necessary to use http2 functionality.

PR #470, opened 2021-04-21
Allow configuration of persistent TCP connections

edevil:
Added 2 new options to configure previously hardcoded
values: max-reuse-tcp-queries and tcp-reuse-timeout. These
allow fine-grained control over how unbound uses persistent
TCP connections to authority servers.

Some authority servers don't deal well with sending multiple
queries over the same TCP connection so configuring these
values is necessary.


wcawijngaards (approved):
These options look good and can be merged.

wcawijngaards:
The code looks good so I merged it. From our discussion, an option for the auth tcp timeout would perhaps also be useful in this problem space and could be helpful for solutions.

iz8mbw:
@edevil @wcawijngaards thanks for that!

oerdnj:
Hi @edevil, could you please expand on:

> Some authority servers don't deal well with sending multiple queries over the same TCP connection so configuring these values is necessary.

Do you have any datapoints and can you describe the behaviour by those authoritative servers?  Either here or [dns-operations](https://lists.dns-oarc.net/mailman/listinfo/dns-operations) would be fine, I guess.

edevil:
Hello @oerdnj

It has been a while since I looked at this, but what I was seeing at the time was that some servers did not respond to subsequent queries on the same TCP socket, but also did not close it. So this caused high latency because Unbound would first need to give up on the tcp socket before a successful query was sent.

oerdnj:
Oh, so it was a **client** issue and not a **server** issue.  That's good to know, thank you, @edevil!

edevil:
Well, I wouldn't consider keeping the TCP connection open and ignoring new queries on that connection a good behavior... :)

These options mitigate the issue on the client side but it seems more like a server issue.

PR #472, opened 2021-04-22
View specific caches.

saksham-m:
Lots to do, but what is working:

1) A default view, called "server". Any query not in a view goes to server view. Therefore every query is technically in a view
2) Local-zones and view-first still working as expected
3) View specific caches. Configurable by
view:
msg-cache-size: 20m
rrset-cache-size: 10m
inside a view clause
Not specifying these will cause the view to inherit the server-view cache.

Feedback and comments very welcome.


wcawijngaards (approved):
The code looks good! Thank you for the patch.

Merge needs to wait a little while the main branch has bug fixes as its priority. Also; the patch applies to 1.13.1 and likely needs some adjustment to fit on the main branch.

saksham-m:
Sounds good. Let me know when you are ready and I can re-factor the code for the master again.

PR #481, opened 2021-05-03
Fix comment in configuration file

paulmenzel:


wcawijngaards (approved):
Yes that probably referred to the trust anchor. The change looks good.

wcawijngaards:
It is referring to the 'auto-trust-anchor-file' statement further down, I can write a better fix for this.

wcawijngaards:
Turned it into a full sentence so that it does not look broken with the license reference after it. Thanks for the report!

PR #487, opened 2021-05-08
ifdef RLIMIT_AS in recently added check

sthen:
Some OS still don't have RLIMIT_AS


wcawijngaards:
Thanks for the patch. This RLIMIT_AS check is a convenience check for operators so if it is omitted that should not harm the server activities.

Merged it into the code repository.

PR #493, opened 2021-05-24
Fix generation of libunbound.pc

Jakker:
According to a [PR at FreeBSD ports](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=256064) a broken libunbound.pc file is installed.

This is the proposed patch for it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -824,7 +824,7 @@ ACX_WITH_SSL
 ACX_LIB_SSL
 SSLLIB="-lssl"
 
-PC_CRYPTO_DEPENDENCY="libcrypto libssl"
+PC_CRYPTO_DEPENDENCY=""
 AC_SUBST(PC_CRYPTO_DEPENDENCY)
 
 # check if -lcrypt32 is needed because CAPIENG needs that. (on windows)


wcawijngaards (changes requested):
We cannot just remove these flags for everyone, that would be wrong. From the link you cite, it seems like it needs to be removed if libcrypto and openssl are from the FreeBSD base system. And it should be present when openssl is from ports or on other systems. Thus also the patch, if deployed in the code repository of unbound, has to detect whether this is the FreeBSD base system openssl, before selecting the value of PC_CRYPTO_DEPENDENCY. In the link cited, the developer seems to suggest putting this diff in the package system of FreeBSD; and I guess that sort of detects what openssl is used; but may also be a Makefile that is used for both base system openssl and ports installed openssl.

wcawijngaards:
However, is a requires line, and not the LIBS line, I misread that. Much less harmful if we remove that for everyone; they may compile without proper requirements, but the linking works fine if they have openssl somehow. So I guess, since it just drops the requirement checks from pkg-config, it won't fail the compile for a whole bunch of systems; so if you want I do not mind applying this change so the pkg-config file, in our contrib/ that is requested.

wcawijngaards (approved):
If it solves matters as people suggest, I do not see compile failure inducing blockers.

PR #496, opened 2021-05-27
Use build system endianness if available, otherwise try to work it out.

banburybill:
The getdns build system provides the endianness, enabling building on
Windows native. This change is a convenience for getdns maintainers.


wcawijngaards:
Thanks for the update. It is nice to keep that software in sync.

Issue #495, opened 2021-05-27
Documentation or implementation of "verbosity" option


he32:
**Describe the bug**
The documentation for the `verbosity` option says:
```
       verbosity: <number>
              The verbosity number, level 0 means no verbosity, only errors.
              Level 1 gives operational information. Level 2 gives detailed
              operational information. Level 3 gives query level information,
              output per query.  Level 4 gives algorithm level information.
              Level 5 logs client identification for cache misses.  Default is
              level 1.  The verbosity can also be increased from the
              commandline, see unbound(8).
```

While looking at an operational problem, I wanted to get "more detailed operational information",
so I set `verbosity: 2` in `unbound.conf`, but to my surprise, I got "query level information" in the
log, like this:

```
May 26 20:01:53 dr2 unbound: [13154:0] info: start of service (unbound 1.13.1).
May 26 20:02:12 dr2 unbound: [13154:0] info: control cmd:  stats
May 26 20:02:19 dr2 unbound: [13154:0] info: resolving r3---sn-ovgq0oxu-5goe.googlevideo.com. A IN
May 26 20:02:19 dr2 unbound: [13154:0] info: priming . IN NS
May 26 20:02:19 dr2 unbound: [13154:0] info: resolving E.ROOT-SERVERS.NET. AAAA IN
May 26 20:02:19 dr2 unbound: [13154:0] info: priming . IN NS
May 26 20:02:19 dr2 unbound: [13154:0] info: resolving G.ROOT-SERVERS.NET. AAAA IN
May 26 20:02:19 dr2 unbound: [13154:0] info: priming . IN NS
May 26 20:02:19 dr2 unbound: [13154:0] info: response for . NS IN
May 26 20:02:19 dr2 unbound: [13154:0] info: reply from <.> 2001:500:2f::f#53
May 26 20:02:19 dr2 unbound: [13154:0] info: query response was ANSWER
May 26 20:02:19 dr2 unbound: [13154:0] info: priming successful for . NS IN
May 26 20:02:19 dr2 unbound: [13154:2] info: resolving init.ess.apple.com. TYPE65 IN
May 26 20:02:19 dr2 unbound: [13154:2] info: resolving init.ess.apple.com. A IN
May 26 20:02:19 dr2 unbound: [13154:2] info: resolving a239.gi3.akamai.net. TYPE65 IN
May 26 20:02:19 dr2 unbound: [13154:2] info: resolving a239.gi3.akamai.net. A IN
[...]

PR #497, opened 2021-05-28
No longer log per-query details under verbosity=2 (VERB_DETAIL).

he32:
Instead, push those to per-query logging under verbosity=3
(VERB_QUERY), to match the expectations set by unbound.conf(5).
This should deal with issue #495.


gthess:
Thanks for this but as commented on #495, this is a documentation issue at the end.

PR #504, opened 2021-06-24
Extended DNS Errors a la RFC8914

wtoorop:
Still TODO:

- [x] Answer [EDE code 4 Forged](https://www.rfc-editor.org/rfc/rfc8914.html#name-extended-dns-error-code-4-f) for local-data answers
- [x] Fix unit tests
- [x] Configure option(s?) whether or not local-zone should result in EDE code inclusion
- [x] Write subroutine to parse packets which are REFUSED caused by an ACL to either return REFUSED with or without EDE 18
- [x] figure out how to do ACL PROHIBITED reply FORMERROR
- [x] Document how the `log-val-level:` config options influences the returned EDE option **text**
- [x] add logic for per zone EDE for RPZ and configurable rpz-do-ede
- [x] add remote-control support for the added configuration options
- [ ] configurable stale answer

For future iterations, we would like to have cached EDEs and add DNSSEC indeterminate with a configuration option,


util/data/msgreply.c
```diff
@@ -988,6 +988,35 @@ int edns_opt_append(struct edns_data* edns, struct regional* region,
...
+		return 0;
+	opt->next = NULL;
+	opt->opt_code = LDNS_EDNS_EDE;
+	opt->opt_len = txt_len + sizeof(uint16_t);
+	opt->opt_data = regional_alloc(region, txt_len + sizeof(uint16_t));
+	if(!opt->opt_data)
+		return 0;
+	sldns_write_uint16(opt->opt_data, (uint16_t)code);
+	if (txt_len)
+		strncpy(opt->opt_data + 2, txt, txt_len);
```
wcawijngaards:
```suggestion
		memmove(opt->opt_data + 2, txt, txt_len);
```
You want to always copy that length of data, so memmove. Not use the broken strncpy behaviour, or have stripped off end with zero terminator, or shorter string with uninitialised memory.

TCY16: resolved in 84da240

daemon/worker.c
```diff
@@ -1460,10 +1581,23 @@ worker_handle_request(struct comm_point* c, void* arg, int error,
...
 					lock_rw_unlock(&e->lock);
+
+					// // stale answer? 
+					// if (worker->env.cfg->serve_expired &&
+					// 	*worker->env.now >= ((struct reply_info*)e->data)->ttl) {
+					// 	// EDE Error Code 3 - Stale Answer
+					// 	EDNS_OPT_LIST_APPEND_EDE(&edns.opt_list_out, worker->scratchpad,
+					// LDNS_EDE_STALE_ANSWER, "");
+					// }
+
```
chantra: I thought I add commented about this earlier, but I don't seem to find that message again.
Is there any plan to have stale answers EDE handled from the get go of the RFC support?

chantra:
Not sure if I forgot to submit my previous comment a couple of weeks ago, or if it got overridden as the diff took shape.

TL;DR I am looking forward this making in into the main branch and was wondering if [EDE stale answer (code 3)](https://www.rfc-editor.org/rfc/rfc8914.html#name-extended-dns-error-code-3-s) was going to be supported from the get go as the current diff has it commented out.
Thanks @TCY16 for tackling this!

TCY16:
Hi @chantra, please don't let the commented-out code dishearten you, this is very much WIP 😄 
Although I can't guarantee that EDE stale answer will be in the eventual release, we'll definitively look at this particular piece of code again!

PR #510, opened 2021-07-15
Don't call a function which hasn't been defined

ndptech:
This code path was causing segmentation faults by trying to call ev->vmt->winsock_tcp_wouldblock even though that was NULL


wcawijngaards:
Merged, looks good. I also added more in depth fixes in the commit that puts ifdefs around windows api event calls, so they happen when on windows.

PR #511, opened 2021-07-23
- Reduce unnecessary linking                                                                                                                                             

### LINKED ISSUE #242: About python support
Hello,
there are two configure options related with python: `--with-pyunbound` and `--with-pythonmodule`. Due to lack of documentation it is unclear what these options are do. Correct me if I'm wrong but AFAIU, the `--with-pythonmodule` means that `unbound` will be build with python scripting support, i.e. one may call python script to "extend" unbound or change the output of running `unbound` application (that's how I understand the examples in `pythonmod/{,doc/}examples`). The `--with-pyunbound` option is opposite: it enables to build the python "library" which allows one to perform some task from (standalone) python script.

If I'm right could someone enlighten me why if I pass options `--without-pythonmodule --with-pyunbound` to build only a (standalone) python module the `libunbound.so*` becomes linked with `libpython`? Is it necessary and without linking `libunbound.so` with `libpython` the python module (files located in `/usr/lib/pythonX.Y/site-packages`, namely `_unbound.so` and `unbound.py`) won't work? Is it possible to build unbound with python module without linking `libunbound.so` with `libpython`?

P.S. I'm using Archlinux and considering to use "hash-slinger" but it requires unbound python modules. I asked to provide python module for unbound package but package maintainer don't like that in this case `libunbound.so` is linked with `libpython` ("... really too big a hard dependency for such as small package as unbound ...") and though I made a workaround for this I would like in first place to know why there are two options and what they are do.

yan12125:
When pyunbound is enabled while pythonmodule is not (i.e., `./configure --without-pythonmodule --with-pyunbound`), only the Python library `_unbound.so` uses Python functions, and main programs (`unbound`, `unbound-anchor`, ...) and `libunbound.so` do not. This patch removes unneeded linking.

Note that I didn't commit re-generated autoconf files as that brings too many unrelated changes to `./configure`. If you wish, I will push updated autoconf files.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -61,6 +61,7 @@ PYTHON_CPPFLAGS=-I. @PYTHON_CPPFLAGS@
 CFLAGS=-DSRCDIR=$(srcdir) @CFLAGS@
 LDFLAGS=@LDFLAGS@
 LIBS=@LIBS@
+PYTHON_LIBS=@PYTHON_LIBS@
 LIBOBJS=@LIBOBJS@
 # filter out ctime_r from compat obj.
 LIBOBJ_WITHOUT_CTIME=@LIBOBJ_WITHOUT_CTIME@
@@ -476,7 +477,7 @@ libunbound/python/libunbound_wrap.c:	$(srcdir)/libunbound/python/libunbound.i un
 
 # Pyunbound python unbound wrapper
 _unbound.la:	libunbound_wrap.lo libunbound.la
-	$(LIBTOOL) --tag=CC --mode=link $(CC) $(RUNTIME_PATH) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -module -avoid-version -no-undefined -shared -o $@ libunbound_wrap.lo -rpath $(PYTHON_SITE_PKG) -L. -L.libs libunbound.la $(LIBS)
+	$(LIBTOOL) --tag=CC --mode=link $(CC) $(RUNTIME_PATH) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -module -avoid-version -no-undefined -shared -o $@ libunbound_wrap.lo -rpath $(PYTHON_SITE_PKG) -L. -L.libs libunbound.la $(PYTHON_LIBS)
 
 util/config_file.c:	util/configparser.h
 util/configlexer.c:  $(srcdir)/util/configlexer.lex util/configparser.h
--- a/configure.ac
+++ b/configure.ac
@@ -699,11 +699,15 @@ if test x_$ub_test_python != x_no; then
       AC_SUBST(PY_MAJOR_VERSION)
       # Have Python
       AC_DEFINE(HAVE_PYTHON,1,[Define if you have Python libraries and header files.])
-      if test -n "$LIBS"; then
-        LIBS="$PYTHON_LDFLAGS $LIBS"
-      else
-        LIBS="$PYTHON_LDFLAGS"
+      if test x_$ub_with_pythonmod != x_no; then
+        if test -n "$LIBS"; then
+          LIBS="$PYTHON_LDFLAGS $LIBS"
+        else
+          LIBS="$PYTHON_LDFLAGS"
+        fi
       fi
+      PYTHON_LIBS="$PYTHON_LDFLAGS"
+      AC_SUBST(PYTHON_LIBS)
       if test -n "$CPPFLAGS"; then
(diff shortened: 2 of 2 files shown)


ArchangeGabriel:
At the same time, there is overlinking happening there:
```
python-unbound W: Unused shared library '/usr/lib/libssl.so.1.1' by file ('usr/lib/python3.9/site-packages/_unbound.so')
python-unbound W: Unused shared library '/usr/lib/libsodium.so.23' by file ('usr/lib/python3.9/site-packages/_unbound.so')
python-unbound W: Unused shared library '/usr/lib/libprotobuf-c.so.1' by file ('usr/lib/python3.9/site-packages/_unbound.so')
python-unbound W: Unused shared library '/usr/lib/libevent-2.1.so.7' by file ('usr/lib/python3.9/site-packages/_unbound.so')
python-unbound W: Unused shared library '/usr/lib/libsystemd.so.0' by file ('usr/lib/python3.9/site-packages/_unbound.so')
python-unbound W: Unused shared library '/usr/lib/libcrypto.so.1.1' by file ('usr/lib/python3.9/site-packages/_unbound.so')
python-unbound W: Unused shared library '/usr/lib/libhiredis.so.1.0.0' by file ('usr/lib/python3.9/site-packages/_unbound.so')
python-unbound W: Unused shared library '/usr/lib/libnghttp2.so.14' by file ('usr/lib/python3.9/site-packages/_unbound.so')
```

**EDIT**: The usual `sed -i -e 's/ -shared / -Wl,-O1,--as-needed\0/g' libtool` workaround works for this though.

yan12125:
> At the same time, there is overlinking happening there:

Many thanks for testing! I looked into it, and found two issues:

* Unbound's Makefile specifies unneeded libraries for `_unbound.so`. I address that issue in the updated patch.
* libtool brings unneeded libraries. For example, using `./configure --with-pythonmodule --with-pyunbound`, build commands for `_unbound.so` are:
```
./libtool --tag=CC --mode=link gcc  -I. -I/usr/include/python3.9 -DSRCDIR=. -g -O2 -flto -pthread  -module -avoid-version -no-undefined -shared -o _unbound.la libunbound_wrap.lo -rpath /usr/lib/python3.9/site-packages libunbound.la -L/usr/lib -L/usr/lib/python3.9 -L. -lpython3.9
libtool: link: rm -fr  .libs/_unbound.la .libs/_unbound.lai .libs/_unbound.so
libtool: link: gcc -shared  -fPIC -DPIC  .libs/libunbound_wrap.o   -Wl,-rpath -Wl,/home/yen/var/unbound/.libs -Wl,-rpath -Wl,/usr/local/lib ./.libs/libunbound.so -L/usr/lib -L/usr/lib/python3.9 -L. -lssl -lcrypto -lpython3.9  -g -O2 -flto -pthread   -pthread -Wl,-soname -Wl,_unbound.so -o .libs/_unbound.so
```
Here `-lssl` and `-lcrypto` do not appear in the first `libtool` command, but the final gcc command includes them. Googling around, besides the `sed` command you've mentioned, I also found discussions about [linking "recursively" by libtool](https://wiki.mageia.org/en/Overlinking_issues_in_packaging#libtool_issues) and [a patch from Debian](https://salsa.debian.org/mckinstry/libtool/-/blob/debian/master/debian/patches/link_all_deplibs.patch) to disable it. I guess that is the cause why `-lssl` and `-lcrypto` are added to `_unbound.so` - both libraries are used by libunbound. Hopefully, with this patch, linking is correct for distros that disable recursive linking without further patching.

/cc @orbea Mind to test this patch on Gentoo with `slibtool`? I found that `$(LIBS)` is added for `_unbound.so` in https://github.com/NLnetLabs/unbound/pull/449. I guess replacing `$(LIBS)` with `$(PYTHON_LIBS)` will not break things on Gentoo, but more testing will be better.

orbea:
This patch seems to compile correctly using slibtool, the current git head (dd254137f1450976b43751248ee208008e65fd42) and `--with-pyunbound`.

I think the only way this could fail with slibtool is if `$(LIBS)` contains any required references that `$(PYTHON_LIBS)` does not also contain which would then fail with `-no-undefined` which GNU libtool silently ignores, but that doesn't seem to be the case.

yan12125:
Thanks a lot for testing and clear explanations!

yan12125:
Rebased to resolve conflicts after 8e538dcaa8df2d0fab8ff3dcf94ac1f972450b66

gthess:
Thanks for this! LGTM and the Python tests pass.

PR #512, opened 2021-07-25
unbound.service.in: upgrade hardening to latest standards

ArchangeGabriel:
Systemd gradually introduced new protection bits, let’s enable them.

This enhance the security score from 3.6 to 3.0. I’ve tested it on two systems, but definitively do not cover all cases, so further testing would be welcome.


wcawijngaards:
Thanks for the upgrade, committed the patch to the repository.

PR #514, opened 2021-07-27
Docker environment for run tests

ziollek:
I've prepared Dockerfile which enable to run tests in isolated environment with all needed tools. I hope it helps newbies like me to verify their PR.

While i was running 
```
make longtest 
```

i noticed some undeterministic behaviour (some tests ie. auth_https.tdir sometimes failed due to connection timeout to petal).
I found some problem in **util/netevent.c** which cause switching to reading phase before send http request due to incomplete ssl_handshake.


.gitignore
```diff
@@ -56,3 +56,4 @@
 /testdata/.perfstats.txt
 /doc/html
 /doc/xml
+.idea
```
gthess: I would keep these (editor specific ignores) in your [local (global) git configuration](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreexcludesFile).
We like to populate the `.gitignore` file only with unbound related files.

util/iana_ports.inc
```diff
@@ -4244,6 +4244,7 @@
 5504,
 5505,
 5506,
+5540,
```
gthess: I would prefer to leave this change to the maintainers.
Updating IANA ports is one of the steps leading to a release.

configure
```diff
@@ -17888,6 +17889,11 @@ if test "`uname`" = "NetBSD"; then
 	NETBSD_LINTFLAGS='"-D__RENAME(x)=" -D_NETINET_IN_H_'
 
 fi
+
+if test "`uname -o`" = "GNU/Linux"; then
+	# splint cannot parse modern c99 header files
+	GCC_DOCKER_LINTFLAGS='-syntax'
+fi
```
gthess: These changes need to go to `configure.ac` instead. `configure` is auto-generated (also) by the maintainers based on that file.

doc/README.tests
```diff
@@ -15,6 +15,14 @@ You need to have the following programs installed and in your PATH.
 * xxd and nc (optional) - for (malformed) packet transmission.
 The optional programs are detected and can be omitted.
 
+You can also use prepared Dockerfile to run tests inside docker based on latest gcc image:
+* build container: docker build -t unbound-tester .
```
gthess: This would need to be updated for the new Dockerfile location.

util/netevent.c
```diff
@@ -1232,6 +1232,13 @@ ssl_handshake(struct comm_point* c)
 		if(want == SSL_ERROR_WANT_READ) {
 			if(c->ssl_shake_state == comm_ssl_shake_read)
 				return 1;
+			/* According to https://www.openssl.org/docs/man1.1.1/man3/SSL_do_handshake.html
+			 * we should repeat handshake - for non blocking BIO
+			 */
+			if(c->ssl_shake_state == comm_ssl_shake_write) {
+			    comm_point_listen_for_rw(c, 0, 1);
+			    return 1;
+			}
```
gthess: With this condition now, the code that follows turns to dead code.
I would rewrite as:
```
if(want == SSL_ERROR_WANT_READ) {
			/* According to https://www.openssl.org/docs/man1.1.1/man3/SSL_do_handshake.html
			 * we should repeat handshake - for non blocking BIO
			 */
			if(c->ssl_shake_state == comm_ssl_shake_write) {
			    comm_point_listen_for_rw(c, 0, 1);
			}
			return 1;
```

The same logic, adjusted, can also be used for the the `SSL_ERROR_WANT_WRITE` case below.

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+WORKDIR /usr/src/unbound
+RUN apt-get update
+# install semantic parser & lexical analyzer
+RUN apt-get install -y bison flex
+# install packages used in tests
+RUN apt-get install -y ldnsutils dnsutils xxd splint doxygen netcat
+# accept short rsa keys, which are used in tests
+RUN sed -i 's/SECLEVEL=2/SECLEVEL=1/g' /usr/lib/ssl/openssl.cnf
+
+CMD ["/bin/bash"]
```
gthess: I would prefer this file to be renamed to `Dockerfile.tests` and also moved under the `contrib/`directory.
Otherwise it gives the impression that it is the main Dockerfile to run unbound.

gthess (changes requested):
Hi, thanks for taking the time for this!

I have some comments for changes before being able to merge.
If something is not clear let me know.

I was hunting an ssl bug in windows and came to the same conclusion for the ssl handshake. May I ask how you stumbled upon this?

util/netevent.c
gthess: We had a better look for this change and we solved our case (https on windows) with commit https://github.com/NLnetLabs/unbound/commit/ca67691092127aafd3c11ead622dcb659959e1bf.
Can you try that one and see if it also solves your use case?

Btw the ssl handshake is (re)tried on both read and write events until the c->ssl_shake_state reaches comm_ssl_shake_none.

util/iana_ports.inc
ziollek: I've reverted this change

.gitignore
ziollek: I've reverted this change

ziollek:
> Hi, thanks for taking the time for this!
> 
> I have some comments for changes before being able to merge.
> If something is not clear let me know.
> 
> I was hunting an ssl bug in windows and came to the same conclusion for the ssl handshake. May I ask how you stumbled upon this?

Thank You for review my PR. I observed problem with handshake while launching longtests in docker, i started digging and found ssl connection was waiting for response despite request wasn't send due to not finished handshake.

I merged actual master to my branch and verified that change ca67691 fixed also above problem - so i've reverted my changes in netevent.c. I've adjusted my code to other suggestions. Please let me know if i can improve something else.

Best regards.

Dockerfile
ziollek: done

util/netevent.c
ziollek: Thanks! This solution resolves also my use cases.

doc/README.tests
ziollek: done

configure
ziollek: done

gthess:
Thanks, looks good to me!
Currently we are in code freeze for the upcoming release but I will merge this right after that.

Something that just came to mind, I am not sure about the portability of the `-E` option to grep. In most man pages I find online I see support for it but we will find out when testing on different platforms for the release. Worst case we convert it back to regular regular expression.

ziollek:
Thanks for information :)

gthess (approved):

PR #517, opened 2021-08-03
#420 breaks the mesh reply list function that need to reuse the dns answer.

dyunwei:
Sorry for my previous change causes another issue.

My previous change that clears the c->buffer in the comm_point_send_reply does resolve the "can't fit qbuffer in c->buffer" issue, but it breaks the mesh reply list function that need to reuse the answer. Since the c->buffer is cleared in the comm_point_send_reply, it cannot be resued again. it means that it is not inappropriate to clear c->buffer in the comm_point_send_reply.

After some investigation, i found it is appropriate to clear c->buffer before use in the http2_query_read_done.

### DIFF
--- a/services/listen_dnsport.c
+++ b/services/listen_dnsport.c
@@ -2477,6 +2477,10 @@ static int http2_query_read_done(struct http2_session* h2_session,
 			"buffer already assigned to stream");
 		return -1;
 	}
+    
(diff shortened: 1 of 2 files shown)


wcawijngaards:
Thanks for the fixup, merged it into the code repository.

PR #522, opened 2021-08-06
- memory management violations fixed

sibeream:
1. SSL CTX is not freed on error
2. No check if region memory was properly allocated


gthess:
Thanks for this!

PR #523, opened 2021-08-11
fix: free() call more than once with the same pointer

Shchelk:
If in the function 'parse_create_msg' failed calls 'reply_info_alloc_rrset_keys' or 'parse_copy_decompress' than in the same place will be called reply_info_parsedelete(*rep, alloc). After that 'reply_info_parsedelete(*rep, alloc)' will be called second time at "reply_info_parse()"


wcawijngaards:
Thank you for the bug fix. That should remove double free issues.

PR #529, opened 2021-08-20
Fix: log_assert does nothing if UNBOUND_DEBUG is undefined

Shchelk:
Static analyzer message: Integer value 'len' obtained from untrusted source at tube.c:374 by passing as 2nd parameter to function 'read' at tube.c:340 without checking its higher bound is used as a loop bound at tube.c:374.

### DIFF
--- a/util/tube.c
+++ b/util/tube.c
@@ -363,7 +363,11 @@ int tube_read_msg(struct tube* tube, uint8_t** buf, uint32_t* len,
 		}
 		d += r;
 	}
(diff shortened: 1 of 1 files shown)


wcawijngaards:
Thank you for the patch! The code does not seem to be at risk, because the messages along this channel are from the internal parts of unbound, apart from bugs, bad lengths are not expected here. Nevertheless, it is good to fix to make sure.

PR #530, opened 2021-08-20
Fix: dereferencing a null pointer

Shchelk:
Static analyzer message: Return value of a function 'reply_info_copy' is dereferenced at dns64.c:923 without checking, but it is usually checked for this function (4/5).


gthess:
Thanks for this!

PR #531, opened 2021-08-20
Fix: passed to proc after free

Shchelk:
Static analyzer message: Pointer 'dp' is passed to a function at iter_hints.c:401 after the referenced memory was deallocated at iter_hints.c:174 by passing as 3rd parameter to function 'hints_insert' at iter_hints.c:398.


wcawijngaards:
Thank you for the fix. The ignored double entry causes the problem here, but it is still good to deallocate, so the fix is the right one.

PR #532, opened 2021-08-20
Fix: buffer overflow bug

Shchelk:
Static analyzer message: Array 'token' of size 65536 bytes passed to function 'rrinternal_parse_rdata' at str2wire.c:679 by passing as 2nd parameter to function 'rrinternal_parse_rdata' at str2wire.c:775, where it is accessed by unacceptable index. This may lead to buffer overflow.


gthess:
Thanks for the fix! This could result to buffer overflow when reading HIP records from string.

PR #538, opened 2021-09-03
Fix subnetcache statistics

tduboys:
Hi all,
Following https://github.com/NLnetLabs/unbound/commit/d69132b9212f95c72fdfb8d0a652f6f58c069a9f
the subnet module was renamed to subnetcache to improve logging, but stats are not displayed correctly.

see https://github.com/NLnetLabs/unbound/blob/master/edns-subnet/subnetmod.c#L864

Tested on 1.13.2 on Debian Stretch. 

The subnet cache worked as expected, the issue was only on statistics (on unbonud-control stats)

### DIFF
--- a/daemon/remote.c
+++ b/daemon/remote.c
@@ -813,7 +813,7 @@ print_mem(RES* ssl, struct worker* worker, struct daemon* daemon,
 	iter = mod_get_mem(&worker->env, "iterator");
 	respip = mod_get_mem(&worker->env, "respip");
 #ifdef CLIENT_SUBNET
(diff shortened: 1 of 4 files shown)


tduboys:
It fix only num.query.subnet and num.query.subnet_cache statistics, not mem.mod.subnet.

wcawijngaards (approved):
Changes look good to me.

wcawijngaards:
Thanks for the additional fixes! That failed string would just make the statistics not work, but not fail compile. Merged it into the code repository.

Issue #501, opened 2021-06-10
[FR] support dohpath key in SVCB records


crolfe:
This a request to implement/support the [Service Binding Mapping for DNS Servers](https://www.ietf.org/id/draft-schwartz-svcb-dns-03.html) RFC, which defines a new `SvcParamKey` (i.e. `dohpath`) for use in SVCB records.  This would be an extension of the [current work to add SVCB support](https://github.com/NLnetLabs/unbound/pull/491) to Unbound.  

**Current behavior**
Once [this PR](https://github.com/NLnetLabs/unbound/pull/491) lands, Unbound will support a list of standard `SvcParamKey`s in a SVCB / HTTPS record.  If an answer contains an unknown SvcParamKey (e.g. `dohpath`) it will replaced by `key65535`, as that is what the RFC says to use for "Invalid key".

```
;ANSWER
resolver.arpa. 300 IN SVCB 1 . alpn="h2"  key65535="/dns-query?{dns}"
```

**Describe the desired feature**
Unbound will recognize `dohpath` as a valid SvcParamKey in SVCB records.  When a client queries for a SVCB record that contains a `dohpath`, this should be present in the answer:

```
;ANSWER
resolver.arpa. 300 IN SVCB 1 . alpn="h2" dohpath="/dns-query?{dns}"
```

**Potential use-case**
* Unbound instances will be able to support [Discovery of Desginated Resolvers](https://datatracker.ietf.org/doc/html/draft-ietf-add-ddr-00), which defines a mechanism for DNS clients to discover encrypted DNS services (e.g. DoH, DoT, DoQ) provided by their dns53 resolver.

**Current Blockers**
* The [SVCB PR](https://github.com/NLnetLabs/unbound/pull/491) is still a draft. I am not sure what the timelines are for getting this ready for review / merged
* An IANA number needs to be allocated for the new `dohpath` SvcParamKey, which to my knowledge, has not happened yet

okanaganrusty:
It looks like `dohpath` will be accepted as parameter key `7`.

![image](https://user-images.githubusercontent.com/10759233/135905846-89f93469-1292-40b0-b33f-9daa2bd295d3.png)

https://www.ietf.org/archive/id/draft-ietf-add-svcb-dns-00.html#name-dohpath

jp-coding:
**Add SVCB dohpath support #739**

jp-coding:
Adressed by #739?

gthess:
[...]

PR #547, opened 2021-10-04
Adds SVCB parameter key 7 (dohpath)

okanaganrusty:
PR to add support for `dohpath` in the SVCB request.

Supporting issue #501

https://datatracker.ietf.org/doc/html/draft-schwartz-svcb-dns-00#section-5.1


wcawijngaards:
Thank you for the code! This looks like it implements the dohpath option for SVCB and HTTPS fine. The document is still a draft, and because we do not want to create a too-early deployment issue, we do not want to ship this at this early stage. It would be best if we waited for discussion at the IETF to complete.

Meanwhile, for experimentation there can be code. The code from the diff of this pullrequest can perhaps help people that want it. Deploy by applying the patch at https://github.com/NLnetLabs/unbound/pull/547.diff and then dohpath should work. Another way, but perhaps later, is to add compile time switches that default to off, that turn on this new feature.

sldns/str2wire.c
```diff
@@ -1495,7 +1500,37 @@ sldns_str2wire_svcbparam_alpn_value(const char* val,
...
+	if (val_len > sizeof(unescaped_dst)) {
+		return LDNS_WIREPARSE_ERR_SYNTAX_INTEGER_OVERFLOW;
+	}
+
+	if (val_len > 255) {
+		return LDNS_WIREPARSE_ERR_SVCB_DOHPATH_KEY_TOO_LARGE;
+	}
+
+	dst_len = sldns_str2wire_svcbparam_parse_copy_unescaped(dst, val, val_len);
+
```
TCY16, line 1527: This write could result in a buffer overflow, as the `rd_len` isn't checked before writing the data. It's missing:
```suggestion
	if (*rd_len < 4 + dst_len)
		return LDNS_WIREPARSE_ERR_BUFFER_TOO_SMALL;
```

testdata/svcb.tdir/svcb.success-cases.zone.cmp
```diff
@@ -8,3 +8,4 @@ s06.success-cases.	3600	IN	HTTPS	0 . ech="aGVsbG93b3JsZCE="
 s07.success-cases.	3600	IN	HTTPS	0 . ech="aGVsbG93b3JsZCE="
 s08.success-cases.	3600	IN	HTTPS	0 . key11="a" key12="a" key13="a" key14="a" key15="a" key16="a" key17="a" key18="a" key19="a" key110="a" key111="a" key112="a" key113="a" key114="a" key115="a" key116="a" key117="a" key118="a" key119="a" key120="a" key121="a" key122="a" key123="a" key124="a" key125="a" key126="a" key127="a" key128="a" key129="a" key130="a" key131="a" key132="a" key133="a" key134="a" key135="a" key136="a" key137="a" key138="a" key139="a" key140="a" key141="a" key142="a" key143="a" key144="a" key145="a" key146="a" key147="a" key148="a" key149="a" key150="a" key151="a" key152="a" key153="a" key154="a" key155="a" key156="a" key157="a" key158="a" key159="a" key160="a" key161="a" key162="a" key163="a"
 s09.success-cases.	3600	IN	HTTPS	0 . alpn="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+s10.success-cases.	3600	IN	HTTPS	0 . alpn="h2" ipv4hint="127.0.0.1" ipv6hint="fe80::1" dohpath="/dns-query?{dns}"
```
TCY16, line 11: This test seems to fail currently. Removing the quotes around the ipv*hints fixes this.
```suggestion
s10.success-cases.	3600	IN	HTTPS	0 . alpn="h2" ipv4hint=127.0.0.1 ipv6hint=fe80::1 dohpath="/dns-query?{dns}"
```

TCY16 (changes requested):
Hi @okanaganrusty,

Thanks again for the code. I found two small changes, but overall it looks very nice :)

jp-coding:
Adressed by #739

gthess:
This seems to be addressed by #739 which is now merged; closing this.

PR #549, opened 2021-10-07
Better positioning of general EDNS option handling

wtoorop:
General EDNS options like PADDING and NSID were only returned with answers from cache or from authoritative zones.
This commit repositions the handling of general EDNS options (from the apply_edns_options() function) so it is better cooperating with handling of EDNS options from modules, and as a result the options are handled for synthesized answers (like local data and id.server queries) too.

@gthess I have some questions about this. Can we discuss?

### DIFF
--- a/daemon/worker.c
+++ b/daemon/worker.c
@@ -449,7 +449,6 @@ answer_norec_from_cache(struct worker* worker, struct query_info* qinfo,
 	 * Then check if it needs validation, if so, this routine fails,
 	 * so that iterator can prime and validator can verify rrsets.
 	 */
-	struct edns_data edns_bak;
 	uint16_t udpsize = edns->udp_size;
 	int secure = 0;
 	time_t timenow = *worker->env.now;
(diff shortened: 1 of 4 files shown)


TCY16 (approved):

wtoorop:
@gthess and @wcawijngaards After discussion with @gthess this is ready to be merged I think.
I will push the merge button (and add a Changelog items for it) if you both approve.

gthess (approved):
I approve!
Also testcases for NSID, padding and keepalive would be nice now for replies from (cache, local-data. etc).

wcawijngaards (approved):
I think it is a hack to have the edns handling in the module callback function as some sort of side effect. But it does fix the issues with EDNS option handling for the separate cases.

What is wrong is that the routine to call the callback should just do that. Now it does more, in reality something that could be fixed by a name change of the function, or something like that. If you think that is better. I approve of the code regardless.

wtoorop:
Thanks @wcawijngaards , I agree. The problem is that the module callbacks also _change_ the options in the edns struct from "incoming" to "outgoing" options.
We would have more flexibility in positioning were final options are applied (for example in `attach_edns_record()` from `util/data/msgencode.c`) when there would be an `options_in` and `options_out` list on the `edns_data` struct instead of a single `options` list for both. WDYT @wcawijngaards @gthess and @TCY16 .

wtoorop:
Closing this in favor of #563

PR #556, opened 2021-10-25
Better positioning of general EDNS option handling; revisited

TCY16:
This implements the discussion we had with @wtoorop, @gthess, and @wcawijngaards on another way to handle EDNS options in #549. To the EDNS struct we add an outgoing options list, so we do not reuse/rewrite the incoming options list.


TCY16:
@wtoorop, @gthess and I decided that we don't pursue this solution anymore so I'm closing this PR

PR #565, opened 2021-11-03
unbound.service.in: Disable ProtectKernelTunables again

Maryse47:
This option was removed in https://github.com/NLnetLabs/unbound/commit/ff8fd0be5c529e7a1b84e8c74426e9c531c0a8f8 but reintroduced in https://github.com/NLnetLabs/unbound/commit/c32b9e4ba95983146eac805719db720f02a64358

Disable it with commentary in hope to prevent slipping it in again.

cc: @ArchangeGabriel

### DIFF
--- a/contrib/unbound.service.in
+++ b/contrib/unbound.service.in
@@ -64,7 +64,8 @@ ProtectClock=true
 ProtectControlGroups=true
 ProtectKernelLogs=true
 ProtectKernelModules=true
-ProtectKernelTunables=true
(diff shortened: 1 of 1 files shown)


Maryse47:
@wcawijngaards could you take a look at this one?

wcawijngaards (approved):
Yes it makes sense to stop this from getting removed and added again to protect, in a cycle.

ArchangeGabriel:
Sorry, I should have looked at the file history. Explicitly disabling is a good idea indeed.

PR #570, opened 2021-11-13
Fix typos

rex4539:


wcawijngaards (approved):
These corrections look correct to me.

wcawijngaards:
Thank you for the spelling fixes. I merged them. The aclocal.m4 file is generated, so I created it again and this removed the spelling fixes, the issues are in the underlying code that is included to make the file, and it is not from the unbound package. The configure file is similar, and I fixed two spelling issues in the source configure.ac for it, and then generated it again, one change is then removed as the text is source in from an underlying source. That could be in libtool, perhaps.

Issue #588, opened 2021-12-15
Unbound 1.13.2 crashes due to p->pc is NULL in serviced_udp_callback


dyunwei:
Below is a part of the call stack:
```
#0  0x00000000004f1cf7 in serviced_udp_callback (c=0x445ca490, arg=0x84ecef40, error=0, rep=0x7ffcfa077c40) at services/outside_network.c:3112
#1  0x00000000004ed10d in outnet_udp_cb (c=0x445ca490, arg=0x44286b50, error=0, reply_info=0x7ffcfa077c40) at services/outside_network.c:1424
```
At frame 0 in the call stack , i printed "struct pending* p" as below:

```
(gdb) p	*p
$2 = {node = {parent = 0x442f84d0, left = 0x0, right = 0x0, key = 0x863e1ca0, color = 1 '\001'}, id = 7761632, addr = {ss_family = 16520,
    __ss_padding = "...", '\000' <repeats 47 times>, __ss_align = 0}, addrlen = 0, pc = 0x0, timer = 0x0, cb = 0x0, cb_arg = 0x0, outnet = 0x0,	sq = 0x0, next_waiting = 0x0,
  timeout = 16,	pkt = 0x71d54668 "j\210", pkt_len = 56}
```
From this display, it can be known that the p->pc is NULL which causes the crash.

Looking into frame 1 in the call stack,  and printed "struct pending* p" and "p->sq" in outnet_udp_cb.
```
(gdb) p	*p
$32 = {node = {parent = 0x766ee0 <rbtree_null_node>, left = 0x766ee0 <rbtree_null_node>, right = 0x766ee0 <rbtree_null_node>, key = 0x7c3d7ea0,	color = 0 '\000'}, id = 16484, addr = {ss_family = 2,
    __ss_padding = "\000\065\300)\242\036", '\000' <repeats 111	times>,	__ss_align = 0}, addrlen = 16, pc = 0x445ca460,	timer = 0x56b6c180, cb = 0x4f1c8b <serviced_udp_callback>, cb_arg = 0x84ecef40,
  outnet = 0x44286b50, sq = 0x84ecef40,	next_waiting = 0x0, timeout = 0, pkt = 0x0, pkt_len = 0}

(gdb) p	*p->sq
[...]

Issue #537, opened 2021-08-31
Unbound 1.13.2 double free() in serviced_node_del() in poor network conditions


Zerg2000:
Unbound is crashing often in poor network conditions. I have backup network connection and when primary fails, then route metric is automatically changed to secondary. Unbound often crashes during that transitions. You could try to use iptables "random" module to try to reproduce that. I have attached config.

Compilation info (I will provide better backtrace next time, it was not full debug build, variables were optimized out):
Kernel: `5.13.12-gentoo`
Version `1.13.2`
Configure line:
```
--prefix=/usr
--build=x86_64-pc-linux-gnu
--host=x86_64-pc-linux-gnu
--mandir=/usr/share/man
--infodir=/usr/share/info
--datadir=/usr/share
--sysconfdir=/etc
--localstatedir=/var/lib
--docdir=/usr/share/doc/unbound-1.13.2
--htmldir=/usr/share/doc/unbound-1.13.2/html
--with-sysroot=/
--libdir=/usr/lib64
--disable-debug
--disable-gost
--disable-dnscrypt
--disable-dnstap
--enable-ecdsa
--disable-subnet
--disable-cachedb
--disable-static
--disable-systemd
--without-pythonmodule
--without-pyunbound
--without-pthreads
--without-libnghttp2
--disable-flto
--disable-rpath
--enable-event-api
--enable-ipsecmod
--enable-tfo-client
--enable-tfo-server
--with-libevent=/usr
--without-libhiredis
--with-pidfile=/run/unbound.pid
--with-rootkey-file=/etc/dnssec/root-anchors.txt
--with-ssl=/usr
--with-libexpat=/usr
Linked libs: libevent 2.1.11-stable (it uses epoll), OpenSSL 1.1.1k  25 Mar 2021
Linked modules: dns64 ipsecmod respip validator iterator
TCP Fastopen feature available
```
Compilation flags:
```
-mtune=native -O2 -pipe -fstack-protector-all -fstack-clash-protection -ggdb
```

Config:
```
server:
    verbosity: 1
    num-threads: 4
    interface: 127.0.0.1
    interface: ::1
    interface: 192.168.0.1
    outgoing-range: 4096
    outgoing-port-permit: 32768-60999
    outgoing-port-avoid: 0-32767
    so-reuseport: yes
    ip-transparent: yes
    max-udp-size: 3072
    msg-cache-size: 64m
    msg-cache-slabs: 4
    num-queries-per-thread: 1024
[...]

Issue #432, opened 2021-02-20
unbound 1.13.0 crash


toralf:
the crash and related  is seen here: https://lkml.org/lkml/2021/2/20/108

Few more info for the package here at a stable Gentoo hhrdened server:
```
#  emerge -qpvO unbound
[ebuild   R   ] net-dns/unbound-1.13.0  USE="dnscrypt ecdsa http2 threads -debug -dnstap -ecs -gost (-libressl) -python -redis (-selinux) -static-libs -systemd -test" PYTHON_SINGLE_TARGET="python3_8 -python3_7 -python3_9" 
```

gthess:
There is not a lot of information about the crash at the link but the current master branch includes a couple of fixes that could lead to random crashes; especially when reloading Unbound.

gthess:
No extra information on this and in the meantime a random crash issue was fixed (#612); closing.

PR #612, opened 2022-01-25
TCP race condition

gthess:
This PR introduces changes to TCP in an attempt to fix race condition bugs.

It relates to #411, #537, #469, #588 and #432.


wcawijngaards (approved):
Good, I see no problems with this.

gthess:
Thanks!

PR #616, opened 2022-01-31
Update ratelimit logic

gthess:
This PR updates the ratelimit logic to work with the newly introduced changes to serviced_query and the logic that sends a query to the network. It also introduces ratelimit-backoff and ip-ratelimit-backoff for an optional more aggressive countermeasure when the limit is reached.

@wcawijngaards: this changes the function signature for send_query. I don't remember if for that change I have to also address something else than what is already addressed here.


testdata/ratelimit.tdir/ratelimit.pre
```diff
@@ -0,0 +1,33 @@
...
+. ../common.sh
+get_random_port 2
+UNBOUND_PORT=$RND_PORT
+STUB_PORT=$(($RND_PORT + 1))
+CONTROL_PORT=$(($RND_PORT + 2))
+echo "UNBOUND_PORT=$UNBOUND_PORT" >> .tpkg.var.test
+echo "STUB_PORT=$STUB_PORT" >> .tpkg.var.test
+echo "CONTROL_PORT=$CONTROL_PORT" >> .tpkg.var.test
+
+# start ldns-testnd
```
wcawijngaards:
```suggestion
# start ldns-testns
```

wcawijngaards (approved):
Apart from default value and spelling fix, the code looks good as is.

gthess:
Thanks!

PR #631, opened 2022-02-18
Replace OpenSSL's ERR_PACK with ERR_GET_REASON

valldrac:
ERR_PACK is internal of OpenSSL and there's no need to filter out by the function name in this case. The comment in the code already suggests to replace it with ERR_GET_REASON.

It also fixes this error when compiling Unbound with BoringSSL:
```
C/C++: .../boringssl/include/openssl/err.h:468:9: note: macro 'ERR_PACK' defined here
C/C++: #define ERR_PACK(lib, reason)                                              \
C/C++:         ^
C/C++: util/netevent.c:1215:55: error: too many arguments provided to function-like macro invocation
C/C++:                 err == ERR_PACK(ERR_LIB_SSL, SSL_F_SSL3_GET_RECORD, SSL_R_HTTP_REQUEST) ||
```


wcawijngaards (approved):
Sure that code looks good and more generic.

wcawijngaards:
Thanks for the patch! Merged it to increase openssl library compatibility.

PR #632, opened 2022-02-20
Match cnames in ipset

scottrw93:
ipset module does not add ip to list when the specified domain is a CNAME


gthess:
Thanks for this! I went ahead and made some further fixes and introduced an ipset test.

Issue #643, opened 2022-03-08
configuring `--with-libunbound-only` causes .pc file not to be installed


FGasper:
**Describe the bug**
When you build and install _just_ libunbound, the .pc file isn’t installed. This makes it unduly complex to link against libunbound.

**To reproduce**
1. Unpack Unbound 1.15.0
2. `./configure --with-libunbound-only --prefix=$(pwd)/mybuild && make install`
3. `find mybuild | grep pc`

**Expected behavior**
The above should show `mybuild/lib/pkgconfig/libunbound.pc`.

**System:**
 - Unbound version: 1.15.0
 - OS: Ubuntu 20
 - `unbound -V` output: n/a

**Additional information**
The pkg-config file does install if you omit `--with-libunbound-only`.

wcawijngaards:
The merged fix should fix the problem. The issue is closed.

PR #644, opened 2022-03-08
Make `install-lib` make target install the pkg-config file.

FGasper:
Issue #643


wcawijngaards (approved):
Changes look good. The finish operation acts on the library binary, so I would keep those lines together. But it works regardless.

wcawijngaards:
Thanks for the fix! Merged it into the repository.

PR #660, opened 2022-04-08
Sha1 runtime insecure

### LINKED ISSUE #656: [FR] Smart deprecation of SHA-1 signatures
**Current behavior**
> Is there a current behavior that the feature relates to?

Yes, DNSSEC verifies names signed by algorithm 5 or 7 by default. It is possible to disable validation on build-time for all names. Our engineers prepared system-wide policy, which attempts to enforce [RFC 9155](https://datatracker.ietf.org/doc/html/rfc9155#section-2) and refuse SHA-1 as valid signature algorithm. Sure, it clashes with [RFC 8624](https://datatracker.ietf.org/doc/html/rfc8624#section-3.1), which in turn requires RSASHA-1 to be enabled. Our crypto team made change to OpenSSL policy, which refuses RSASHA-1 signatures verification by default. That would break any RSASHA-1 domains resolution.

Current possible mode is to always disable SHA-1 at build time or leave it enabled. But this behaviour depends on runtime changes to the system.

> If yes, would you wish the current behavior to change?

It seems current way is to check algorithm support before validation starts. In our case, it would be better to leave algorithm support check on OpenSSL library. Problem is there does not seems to be any public API, where it may indicate before validation that RSASHA-1 is not available in runtime.

I would like unbound to be able to fall back from secure handling of signature in ``verify_canonrrset`` function to insecure record, if ``EVP_VerifyFinal()`` returns code indicating used digest function is not acceptable. I would like possiblity to make the record insecure, without ad flag. But avoid SERVFAIL status code and provide result just like the record were not signed at all. It seems some EVP error codes might indicate there is nothing wrong with signature itself, but it is not available due to policy. There seems no alternative 

**Potential use-case**
RHEL9 and CentOS Stream 9 will make ``EVP_VerifyFinal()`` _fail_ in crypto-policy *DEFAULT* mode. But if its mode is changed runtime by administrator to *DEFAULT:SHA1*, it makes EVP_VerifyFinal() _pass_ even on RSASHA-1 signatures. I would like single binary build to handle both modes.

- if SHA1 is disabled, produce INSECURE results without ad bit
- if SHA1 is enabled, produce SECURE results with ad bit set
- avoid SERVFAIL status if verification failure is related to system-wide policy, not wrong hash or bad key

pemensik:
Attempt to solve issue #656

Fallback from bogus to insecure, if crypto library returned specific error during verification.

### DIFF
--- a/validator/val_secalgo.c
+++ b/validator/val_secalgo.c
@@ -97,6 +97,23 @@ log_crypto_error(const char* str, unsigned long e)
 	log_err("%s crypto %s", str, buf);
 }
 
+/**
+ * Output a libcrypto openssl error to the logfile as a debug message.
+ * @param level: debug level to use in verbose() call
+ * @param str: string to add to it.
+ * @param e: the error to output, error number from ERR_get_error().
+ */
+static void
+log_crypto_verbose(enum verbosity_value level, const char* str, unsigned long e)
+{
+	char buf[128];
+	/* or use ERR_error_string if ERR_error_string_n is not avail TODO */
+	ERR_error_string_n(e, buf, sizeof(buf));
+	/* buf now contains */
+	/* error:[error code]:[library name]:[function name]:[reason string] */
+	verbose(level, "%s crypto %s", str, buf);
+}
+
 /* return size of digest if supported, or 0 otherwise */
 size_t
 nsec3_hash_algo_size_supported(int id)
@@ -668,10 +685,11 @@ digest_ctx_free(EVP_MD_CTX* ctx, EVP_PKEY *evp_key,
 }
 
 static enum sec_status
-digest_error_status(void)
+digest_error_status(const char *str)
 {
-#ifdef EVP_R_INVALID_DIGEST
 	unsigned long e = ERR_get_error();
+	log_crypto_verbose(VERB_QUERY, str, e);
+#ifdef EVP_R_INVALID_DIGEST
 	if (ERR_GET_LIB(e) == ERR_LIB_EVP &&
 	    ERR_GET_REASON(e) == EVP_R_INVALID_DIGEST)
 		return sec_status_indeterminate;
@@ -762,15 +780,16 @@ verify_canonrrset(sldns_buffer* buf, int algo, unsigned char* sigblock,
 	}
 #ifndef HAVE_EVP_DIGESTVERIFY
 	if(EVP_DigestInit(ctx, digest_type) == 0) {
-		enum sec_status sec = digest_error_status();
-		verbose(VERB_QUERY, "verify: EVP_DigestInit failed");
+		enum sec_status sec;
+		sec = digest_error_status("verify: EVP_DigestInit failed");
 		digest_ctx_free(ctx, evp_key, sigblock,
 				dofree, docrypto_free);
 		return sec;
 	}
 	if(EVP_DigestUpdate(ctx, (unsigned char*)sldns_buffer_begin(buf), 
 		(unsigned int)sldns_buffer_limit(buf)) == 0) {
-		verbose(VERB_QUERY, "verify: EVP_DigestUpdate failed");
+		log_crypto_verbose(VERB_QUERY, "verify: EVP_DigestUpdate failed",
+				   ERR_get_error());
 		digest_ctx_free(ctx, evp_key, sigblock,
 				dofree, docrypto_free);
 		return sec_status_unchecked;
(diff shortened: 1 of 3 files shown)


pemensik:
Any comments for this change?

gthess:
As I understand this is also related to FIPS. Why not just do something similar as #671 with detecting FIPS mode for the digest? Then we won't have to rely on the openssl error code for detection (which I cannot find proper documentation for its actual meaning and the stability of its use).

pemensik:
The reason for it is simple. It does happen on CentOS 9 Stream and RHEL 9 in default installation. Unlike PR #671 this does not fail only when FIPS mode is enabled. It fails also when DEFAULT crypto-policy is set. Which is preselected policy on mentioned distributions.

Of course it would fail in FIPS mode too. But it is more general, because it has to be. There's a note about that in [Considerations in adopting RHEL 9](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/considerations_in_adopting_rhel_9/assembly_security_considerations-in-adopting-rhel-9#ref_considerations-security-crypto_changes-to-security). And related [release note mention](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/9.0_release_notes/known-issues#known-issue_infrastructure-services) that unbound package build currently disables all SHA-1 validation in all cases. This change would allow secure resolution of SHA-1 base name in legacy policies, but at least insecure resolution in default policy or FIPS mode.

There is pull request to enable [simple check to openssl](https://github.com/openssl/openssl/issues/17662), but until that is present in openssl, it is not that simple.

Also it is good to note, it would affect both our official packages and also unofficial builds, as long as they link to openssl.

pemensik:
How to check this with simple steps:
```
# podman run -ti quay.io/centos/centos:stream9
# dnf install -y unbound
# unbound-host -rvD ietf.org
ietf.org has address 50.223.129.194 (insecure)
ietf.org has IPv6 address 2001:559:c4c7::100 (insecure)
ietf.org mail is handled by 0 mail.ietf.org. (insecure)
# dnf install -y bind-utils
# delv ietf.org
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; validating ietf.org/DNSKEY: no valid signature found
;; insecurity proof failed resolving 'ietf.org/DNSKEY/IN': 8.8.4.4#53
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; validating ietf.org/DNSKEY: no valid signature found
;; no valid RRSIG resolving 'ietf.org/DNSKEY/IN': 8.8.8.8#53
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; EVP_VerifyFinal failed (verify failure)
;; error:03000098:digital envelope routines::invalid digest:crypto/evp/pmeth_lib.c:959:
;; validating ietf.org/DNSKEY: no valid signature found
;; no valid RRSIG resolving 'ietf.org/DNSKEY/IN': 10.0.2.3#53
;; broken trust chain resolving 'ietf.org/A/IN': 10.0.2.3#53
;; resolution failed: broken trust chain
# update-crypto-policies --set DEFAULT:SHA1
# delv ietf.org
; fully validated
ietf.org.		179	IN	A	50.223.129.194
ietf.org.		179	IN	RRSIG	A 5 2 1800 20230512164317 20220512154550 40452 ietf.org. ujh5zmVnBEGATtJu4Or8Aw4Y2mY2d9AFZLHj2qzgD5pDMyfnUeVRIHui IyEhzAHS7sSVOoBg9gXYr//Ol+i0TWc2UTTXwDwZ7iCzSiOdkxSA+CWJ LI4biOi1CR2ypjQAiuOzVRDCOUD1b8zE9z8PUhdWo0J271QqrQ7jMUzZ CTl6mnpgK0eLiY4Yuf1TbE6F7Adl5Yg6YQYq+OBa+1Aeaso/n+jcSgm2 nkHiLTVqUfj3xrCn/KfFTyPi1q3Gb2wnqZQkRBrYvnaoHdRmt1fzw7sf furzw4MbWkCu+jOaVVO74PQ8sCmEABcCjP/3czEVQ6+F4rmpTakrI/x3 QSYFmQ==
# echo "Now unbound-host could return secure result with this change"
```
Replace podman with docker if you wish.

PR #664, opened 2022-04-15
Add prefetch support for subnet cache entries

tilan7663:
Currently, only the entries in the global cache are being prefetched when prefetch is enable. However for queries with ECS option set, the results are being stored in subnet cache instead global cache which bypass the existing prefetch logic.

I think @gthess attempted to solve similar issue [previously](https://github.com/NLnetLabs/unbound/commit/abbc3305d49c93bfb3ac56a8eb474e7f41f6c1c3), but unfortunately the patch doesn't seem to work when we tested it. This PR follows a similar approach, but instead building the prefetch logic on top of the global cache, the prefetch is only applied to subnet cache queries.

Happy to chat more if people think there are better alternatives to implement this feature. Thanks. 

Signed-off-by: Tian Lan <tian.lan@twosigma.com>

### DIFF
--- a/edns-subnet/subnetmod.c
+++ b/edns-subnet/subnetmod.c
@@ -97,8 +97,8 @@ subnet_new_qstate(struct module_qstate *qstate, int id)
 }
 
 /** Add ecs struct to edns list, after parsing it to wire format. */
-static void
-ecs_opt_list_append(struct ecs_data* ecs, struct edns_option** list,
+void
+subnet_ecs_opt_list_append(struct ecs_data* ecs, struct edns_option** list,
 	struct module_qstate *qstate)
 {
 	size_t sn_octs, sn_octs_remainder;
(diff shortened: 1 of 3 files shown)


gthess:
Hi, the branch you are referring to is an attempt to fix an issue where prefetch queries (from the global cache) will not follow any ECS logic. Prefetch logic was never part of the subnet cache but I see this PR tries to change that!
I'll try to merge my previous branch and then happy to have a look here.

tilan7663:
Thanks @gthess , I will look into the test failures, the tests did pass locally for me before I opened the PR.

gthess:
It's because the tests are now ran with `make test` (because of the rename; `.crpl`s run conditionally based on compiled options as part of the 03-testbound tests). That works fine for you where you compile with subnet but not for the default case.

tilan7663:
oh I see, so I should rename the tests names back to *.crpl.

PR #668, opened 2022-04-24
Set IP_BIND_ADDRESS_NO_PORT on outbound tcp sockets

crrodriguez:
When bound to a local address the kernel does not know if the socket
will listen() or connect() and must reserve a port inmediately after
bind() effectively limiting the random port range to ~32k.
when IP_BIND_ADDRESS_NO_PORT is set, the kernel delays source port
allocation until the time the complete 4-tuple is known allowing
a much larger number of combinations


gthess:
Thanks for this! I relocated and made the code more portable.

crrodriguez:
> Thanks for this! I relocated and made the code more portable.

ok, I noticed the code already needs c99 so i didnt bothered with the c89 version.

PR #677, opened 2022-05-12
Allow using system certificates not only on Windows

pemensik:
OpenSSL has a way to load default file. That file might contain usable
certificates to verify common connections. Allow similar trust as on
windows and leave it on openssl package to provide sane defaults.

Also provide use-system-cert alias, because it is not windows specific
anymore.


wcawijngaards (approved):
Looks good. Could be useful if that is configured to add good contents.

PR #683, opened 2022-05-26
Add upstream cookie support

TCY16:
- [x] Create .rpl test with required changed in testbound
- [x] Make cookies privacy aware (i.e. exclusive per source IP address)
    - [ ] Make test for privacy aware cookies
- [x] Make upstream cookies configurable


services/cache/infra.c
```diff
@@ -383,12 +388,24 @@ static void
...
 	time_t timenow)
 {
-	struct infra_data* data = (struct infra_data*)e->data;
+	int i;
+	struct infra_data* data;
+	uint8_t cookie[8] = {0,0,0,0,0,0,0,0};
+
+	for (i = 0; i < 8; i++) {
+		cookie[i] = ub_random_max(infra->random_state,
+			255); // @TODO is 255 correct? macro-ify it?
```
wcawijngaards: It needs to be 256.
```suggestion
			256);
```

services/cache/infra.c
```diff
@@ -685,6 +706,119 @@ infra_edns_update(struct infra_cache* infra, struct sockaddr_storage* addr,
...
+
+	data = (struct infra_data*) e->data;
+
+	if(needtoinsert) {
+		slabhash_insert(infra->hosts, e->hash, e, e->data, NULL);
+	} else {
+		lock_rw_unlock(&e->lock);
+	}
+
+	return &data->cookie;
```
wcawijngaards: Here data is returned that is not locked, but it points into the entry in the cache's data element. That is a race condition, if the entry->data->cookie is accessed or modified.

services/cache/infra.c
```diff
@@ -685,6 +706,119 @@ infra_edns_update(struct infra_cache* infra, struct sockaddr_storage* addr,
...
+	struct lruhash_entry* e = infra_lookup_nottl(infra, addr, addrlen,
+		name, namelen, 1);
+	struct infra_data* data;
+
+	/* cookie length verification should be checked and handled by caller */
+	assert(cookie->opt_len == 24);
+
+	/* the client cookie was set on the outgoing upstream, so the entry
+	 * must exists here */
+	assert(e);
```
wcawijngaards: This is not true, because once the reply comes in, the entry may have fallen out of the cache. By the cache getting full and the item dropping out of it, or potentially another thread has edited the contents. So the item could not be there any more. Insert it anew, or ignore it, or something like that.

services/cache/infra.h
```diff
@@ -52,6 +52,42 @@
 struct slabhash;
 struct config_file;
 
+/* COOKIE @TODO move this to correct spot */
+
+/**
+ * The actual EDNS cookie data
+ */
+union edns_cookie_data {
```
wcawijngaards: This does not work, I think, the 24 byte is not the same length as the struct components underneath. The struct members get structure alignment. And the members of that struct are not guaranteed to be at the correct spots in the 24 byte sequence, I believe.

services/cache/infra.h
```diff
@@ -318,6 +359,40 @@ int infra_edns_update(struct infra_cache* infra,
...
+/**
+ * Find and return the cookie from the infra cache data. Creates an entry in
+ * the cache if there isn't one.
+ * @param infra: infrastructure cache.
+ * @param addr: host address.
+ * @param addrlen: length of addr.
+ * @param name: name of zone
+ * @param namelen: length of name
+ * @param timenow: what time it is now.
+ * @return: struct ends_cookie*
```
wcawijngaards:
```suggestion
 * @return: struct edns_cookie*
```
Spelling fixed.

testcode/testpkts.c
```diff
@@ -1499,6 +1516,92 @@ match_ednsdata(uint8_t* q, size_t qlen, uint8_t* p, size_t plen)
...
+		walk_query_len = 0;
+		log_err("no edns found");
+	}
+
+	/* class + ttl + rdlen = 8 */
+	if(walk_query_len <= 8) {
+		verbose(3, "No edns opt, no cookie");
+		return 0;
+	}
+
```
wcawijngaards, line 1542: You need an additional if-statement on walk_query_len < 12, and if so, it is too short for the read_uint16 you do further on.

testcode/testpkts.c
```diff
@@ -1499,6 +1516,92 @@ match_ednsdata(uint8_t* q, size_t qlen, uint8_t* p, size_t plen)
...
+	walk_query += 8;
+	walk_query_len -= 8;
+
```
wcawijngaards, line 1576: An additional if-statement, that walk_query_len < 4 means it is too short for the read_uint16 further on.

testcode/testpkts.c
```diff
@@ -1499,6 +1516,92 @@ match_ednsdata(uint8_t* q, size_t qlen, uint8_t* p, size_t plen)
...
+	if (sldns_read_uint16(walk_query) != 10 /* LDNS_EDNS_COOKIE */) {
+		verbose(3, "EDNS option is not a cookie");
+		return 0;
+	}
+	if (sldns_read_uint16(walk_query+2) != 24) {
+		verbose(3, "EDNS cookie is not 24 bytes, so not a correct complete cookie");
+		return 0;
+	}
+
```
wcawijngaards, line 1591: You need an additional if-statement that walk_query_len allows for that 24 length more data, and if shorter, return failure.

testcode/testpkts.c
```diff
@@ -1690,6 +1803,120 @@ adjust_packet(struct entry* match, uint8_t** answer_pkt, size_t *answer_len,
...
+		if (walk_query_len > 0 && walk_response_len > 0) {
+			/* create space for the cookie, as the length of the
+			 * rest of the response is 0 */
+
+			/* depending on the incoming cookie, add the server cookie
+			 * or copy the complete cookie to the response */
+			if (sldns_read_uint16(walk_query+2) == 8) {
+				/* copy the EDNS client cookie from the query
+				 * packet to the response */
+				memcpy(walk_response, walk_query, 12);
```
wcawijngaards: memcpy needs aligned pointers, and these are not, they are packet offsets. memmove can copy for these kinds of possibly unaligned places, also for memcpy calls on walk_response that you write further on.

testcode/testpkts.c
```diff
@@ -1690,6 +1803,120 @@ adjust_packet(struct entry* match, uint8_t** answer_pkt, size_t *answer_len,
...
+				/* add the server cookie to the client cookie to make it
+				 * 'complete'. we fake the siphash specified in RFC9018
+				 * by hardcoding the server cookie */
+				memcpy(walk_response+12, hardcoded_server_cookie, 16);
+
+				/* update the RDLEN and OPTLEN */
+				sldns_write_uint16(rdlen_ptr_response, 28);
+				sldns_write_uint16(walk_response+2, 24);
+
+				reslen = origlen + 28;
```
wcawijngaards: reslen was already incremented by 28 at line 1818, I do not think you have to do it again.

testcode/unitmain.c
```diff
@@ -469,11 +469,12 @@ infra_test(void)
 	struct infra_key* k;
 	struct infra_data* d;
 	int init = 376;
+	struct ub_randstate* rnd = ub_initstate(NULL);
```
wcawijngaards, line 472: Put at line 531, at the end of this function, `ub_randfree(rnd);` to delete it.

wcawijngaards (commented):
So I noticed some debug and todo comments and debug log prints, and made some suggestions in the items. The code looks nice. The .rpl file does not exercise the outside_network code, but it is not that big a part of the code.

PR #695, opened 2022-06-13
Zone parsers blank line issue

Philip-NLnetLabs:


wcawijngaards:
A quick initial look, could you also see if the function sldns_bget_token_par, which is further down the same file with similar contents, needs similar fixes for it. Also could you change the unit test so that it does not fire queries to the internet, which I think it does for example.com. Create like a stub-zone that points nowhere, to 127.0.0.1 to an unused port number for it. Or have ldns-testns listening on that port number to catch that query. I think it is neater to not fire actively queries to the network for the tests.

Philip-NLnetLabs:
Do you know how to trigger sldns_bget_token_par? I'll change the upstream to 127.0.0.1

wcawijngaards:
The various string conversions. From `sldns_str2wire_rr_buf` and `sldns_str2wire_rr_question_buf` that are used for like auth zone and local zone, local-data. And for auth zone https downloads. And for the parsing of trust anchors and trust anchor config.

Philip-NLnetLabs:
I fixed the existing test and added a new test that should exercise sldns_bget_token_par

PR #703, opened 2022-06-22
Fix use after free issue with edns options (https://github.com/NLnetL…

Philip-NLnetLabs:
New version


gthess (approved):
LGTM! Now the outgoing ECS option per upstream is properly stored in the serviced_query region.

Issue #713, opened 2022-07-14
infra-keep-probing is causing problems to fast server selection


hunts:
**Describe the bug**
Recently we are tuning our Unbound fleet's performance. To reduce latency for resolving a domain not in the cache, we started serving root zone locally via auth_zone. But we still see large latency, and then we started monitoring performance of TLD's nameservers.

We are seeing, even with fast server selection(num = 3), Unbound does not pick the fast servers. Note, if we do a reload, it is picking the expected servers, but when the infra-host-ttl is reached, Unbound starts behaving like it does not know which one is faster any more.

After doing some experience, I've identified the root cause, a bug in the infra-keep-probing logic, at https://github.com/NLnetLabs/unbound/blob/055f5e68a3b200a40a9572edd1c5b28a2ae29a9c/services/cache/infra.c#L721-L736

When the infra cache entry expired, the host get huge penalty because it will be treated as unresponsive server (USEFUL_SERVER_TOP_TIMEOUT-1000) even it is the fastest one, and thus won't be selected as an upstream candidate because of the logic at
https://github.com/NLnetLabs/unbound/blob/055f5e68a3b200a40a9572edd1c5b28a2ae29a9c/iterator/iter_utils.c#L407-L415
and at
https://github.com/NLnetLabs/unbound/blob/055f5e68a3b200a40a9572edd1c5b28a2ae29a9c/iterator/iter_utils.c#L590-L596

While if the entry does not exist at all, it's rtt is default to 376 (int UNKNOWN_SERVER_NICENESS = 376). Effectively, it means a host with an expired latency data is much worst then an unknown host.
 

**To reproduce**
1. Update server config: `infra-keep-probing: yes`, `fast-server-permil: 500`, `fast-server-num: 3`, `infra-cache-numhosts: 10000`, and `infra-host-ttl: 600`.
2. Restart the service (`sudo systemctl restart unbound.service`0 to get into a clean state.
3. Check `sudo /usr/sbin//unbound-control -c /etc/unbound/unbound.conf stats`, and dump_infra for later reference.
4. Run `for x in {1..200}; do dig @you_ip -p you_port $x.com \+short; done` to fill infra cache
5. Check stats again `sudo /usr/sbin/unbound-control -c /etc/unbound/unbound.conf stats`, and dump_infra to find infra cache of .com's nameserver hosts
[...]

PR #714, opened 2022-07-14
Avoid treat normal hosts as unresponsive servers

hunts:
This is a fix for issue #713

When infra-keep-probing is on, all hosts with expired entries were treated as
unresponsive servers and thus causing problems (see #713).

This commit change that, so that normal hosts with expired entries are treated
as unknown servers.

### DIFF
--- a/services/cache/infra.c
+++ b/services/cache/infra.c
@@ -721,21 +721,20 @@ infra_get_lame_rtt(struct infra_cache* infra,
 			else	*rtt = USEFUL_SERVER_TOP_TIMEOUT-1000;
 		}
 	}
+	/* expired entry */
 	if(timenow > host->ttl) {
-		/* expired entry */
+		lock_rw_unlock(&e->lock);
+
 		/* see if this can be a re-probe of an unresponsive server */
 		/* minus 1000 because that is outside of the RTTBAND, so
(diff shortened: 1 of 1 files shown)


wcawijngaards (commented):
The change looks okay for the timeout mechanism. The lock changes are wrong, it has to hold lock e a little longer.

wcawijngaards:
The locks fixed in commit https://github.com/NLnetLabs/unbound/commit/33bd49af8140e6744064a74eaf1c369ebed86610 . Thanks for the patch!

hunts:
Thanks for merging it. Our internal patch is exact same as the final version.

I thought the lock changes just made the code nicer and they were identical. Do you mind explain a little bit about the reason the lock should be hold a bit longer? This is my first time look into Unbound code, so would like to learn more about that.

wcawijngaards:
The host->rtt.rto that is used in the if statement in line 731 needs the lock on e. At line 699 it can seen that it is a part of e, e->data.

hunts:
> The host->rtt.rto that is used in the if statement in line 731 needs the lock on e. At line 699 it can seen that it is a part of e, e->data.

Ah, that make sense, I missed that. Thanks again!

PR #718, opened 2022-07-16
Introduce infra-cache-max-rtt option to config max retransmit timeout

### LINKED ISSUE #717: [FR] Make RTT_MAX_TIMEOUT configurable
**Current behavior**
`RTT_MAX_TIMEOUT` is currently hard coded to 120,000 milliseconds, i.e. 2 minutes. Which IMO is too large in our network, we'd like to be able to decrease this value. When adding the new option, we can default the value to the current 120,000 so that the current behavior is not changed.

Add a new server option `infra-cache-max-rtt` which is similar to the existing `infra-cache-min-rtt`, but for configuring the upper limit of retransmitting timeout.

Since `USEFUL_SERVER_TOP_TIMEOUT` and `BLACKLIST_PENALTY` derives from `RTT_MAX_TIMEOUT`, when the latter one is overwritten by the config, the former two should be overwritten along with it.

**Potential use-case**
Allow for customizing `RTT_MAX_TIMEOUT` as needed.

hunts:
Added the option and let it default to 120 seconds so that it won't change current behavior.

Related-to #717


wcawijngaards (approved):
These changes look good. The default is what used to happen, so it does not change things for existing users.

Issue #721, opened 2022-07-23
[FR] integrated NAT64 translation for outbound packets (e.g. when DNS64 in use)


eqvinox:
**Current behavior**
This is loosely related to the `dns64` module (and its `dns64-prefix` config option). The existing behavior of this module should *not* change, it's really an independent feature, just in the same ballpark area.

**Describe the desired feature**
When unbound is running in a NAT64/DNS64 scenario, it may not actually have IPv4 connectivity itself (i.e. be behind the same NAT64.) In this case, any domain that has solely IPv4 NS will fail to resolve, since unbound is trying to reach it over IPv4.

An option to make unbound internally translate its outgoing packets from IPv4 to the NAT64 prefix would make this work nicely. This would be optional behavior enabled with a switch somewhere; it could use the `dns64-prefix` option as default, or have its own prefix setting.

(I checked out the unbound source and looked around, but the code base is too complex for me to figure out how to add this feature, sorry.)

**Potential use-case**
The use case for this is running a validating recursor on devices inside a NAT64 domain. Might be a larger network with one fat NAT64 gateway but with a desire for distributing/splitting out recursors. Might be a laptop with a local recursor.

agowa:
Ran into this issue today too. Thanks for the PR, hope it gets merged soon.

gthess:
#722 solved this; closing.

PR #722, opened 2022-07-23
NAT64 support

eqvinox:
This implements #721.  ~It's not really polished but does work.~

~TODO:~

- ~clean up code formatting~
- ~make NAT64 prefix configurable instead of hardcoding it~
- ~tests~

All done, maybe? :laughing:


eqvinox:
Sidenote: I'm perfectly happy with anyone rewriting, mangling, whatever else, this code in order to get it merged. I'll try to respond to comments/change requests/… but if it's something you can just fix in a few minutes, please, go ahead!

iterator/iterator.c
```diff
@@ -2811,6 +2812,32 @@ processQueryTargets(struct module_qstate* qstate, struct iter_qstate* iq,
...
+	if (ie->use_nat64 && real_addr.ss_family == AF_INET) {
+		struct sockaddr_in *sin = (struct sockaddr_in *)&target->addr;
+		struct sockaddr_in6 *sin6;
+
+		real_addr = ie->nat64_prefix_addr;
+		real_addrlen = ie->nat64_prefix_addrlen;
+
+		/* - NAT64 prefix length really doesn't matter at all,
+		 *   it's just a bunch of zero padding
+		 * - config validation requires the NAT64 prefix to be /96
```
eqvinox: Whoops, I completely forgot that RFC6052 actually specifies a different encoding for the prefix lengths. Fixing that in a minute.

eqvinox:
Fixed non-/96 prefixes & added a test using a /48.

momoka0122y:
When this feature is used prefer-ip6 shoud be changed to  "prefer-ip6: yes" instead of the default.
Or else it may synthesize and IPv4 address even when there is an IPv6 address.

gthess:
@momoka0122y in principle you are right and I was thinking of implicitly enabling the `prefer-ip6` option but on second thought I would leave it to the user to explicitly use that.
Unbound keeps track of DNS RTT values to the different upstream IPs and it may be the case that a synthesized IPv4 connection is more favorable than an IPv6 one.
I will add your remark in the documentation though.

agowa:
That sounds suspiciously like RFC 8305 behavior (Sections 3 and 4). https://www.rfc-editor.org/rfc/rfc8305#section-3

And section 7 would explain how to handle NAT64/DNS64 in that context.

gthess (approved):
LGTM!
Some minor fixes, formatting and refactoring that I will address on a follow up commit.
Thanks for this!

PR #739, opened 2022-08-15
Add SVCB dohpath support

TCY16:


sldns/wire2str.c
```diff
@@ -1139,11 +1141,34 @@ static int sldns_wire2str_svcparam_ech2str(char** s,
...
+	uint16_t i;
+
+	assert(data_len > 0); /* Guaranteed by sldns_wire2str_svcparam_scan */
+
+	w += sldns_str_print(s, slen, "=\"");
+
+	/* RC6570#section-2.1 specifies that the '\' (and other non-letter
+	 * characters in the URI) are "intended to be copied literally" */
+	for (i = 0; i < data_len; i++) {
+		// @TODO do a check like isprint()?
```
wcawijngaards: But the parse routine uses sldns_parse_char for the values, and that unescapes every character. Also I think it is prudent to do an isprint check, so that the output is printable characters.

testdata/svcb.tdir/svcb.success-cases.zone.cmp
```diff
@@ -8,3 +8,5 @@ s06.success-cases.	3600	IN	HTTPS	0 . ech="aGVsbG93b3JsZCE="
...
 s08.success-cases.	3600	IN	HTTPS	0 . key11="a" key12="a" key13="a" key14="a" key15="a" key16="a" key17="a" key18="a" key19="a" key110="a" key111="a" key112="a" key113="a" key114="a" key115="a" key116="a" key117="a" key118="a" key119="a" key120="a" key121="a" key122="a" key123="a" key124="a" key125="a" key126="a" key127="a" key128="a" key129="a" key130="a" key131="a" key132="a" key133="a" key134="a" key135="a" key136="a" key137="a" key138="a" key139="a" key140="a" key141="a" key142="a" key143="a" key144="a" key145="a" key146="a" key147="a" key148="a" key149="a" key150="a" key151="a" key152="a" key153="a" key154="a" key155="a" key156="a" key157="a" key158="a" key159="a" key160="a" key161="a" key162="a" key163="a"
 s09.success-cases.	3600	IN	HTTPS	0 . alpn="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+_dns.doh.example.	7200	IN	SVCB	1 doh.example. alpn="h2" dohpath="/dns-query{?dns}"
+_dns.doh.example.	7200	IN	SVCB	1 doh.example. alpn="h2" dohpath="/dns-query{?dns}"
```
wcawijngaards: The diff tool says there is no newline at the end of the file here. Because that gave you issues with NSD, here it is too.

wcawijngaards (approved):
I would prefer an is printable character check on the output, so even if the input is binary, the output is printable text. Otherwise, it looks fine.

testdata/svcb.tdir/svcb.success-cases.zone.cmp
TCY16: My editor says there is a newline at the end of this file. Also, I tested both cases with `readzone` and the Unbound parser accepts both!

sldns/str2wire.c
```diff
@@ -1511,7 +1516,41 @@ sldns_str2wire_svcbparam_alpn_value(const char* val,
...
+	val_len = strlen(val);
+
+	if (*rd_len < 4 + val_len) {
+		return LDNS_WIREPARSE_ERR_BUFFER_TOO_SMALL;
+	}
+
+	/* draft-ietf-add-svcb-dns-06#section-5.1
+	 * The URI Template MUST contain a "dns" variable
+	 */
+	if (!(strstr(val, "?dns"))) {
```
gthess: This does not ensure that the "?dns" value is part of the templating i.e., part of `{...}`.

gthess (changes requested):
I believe you miss the MUST in section 5.1
>If the "alpn" SvcParam indicates support for HTTP, "dohpath" MUST be present.

I would also like to see failure-cases testing, like the above and ?dns missing from the dohpath.

Also more happy testing for printing non-printable characters?

sldns/wire2str.c
```diff
@@ -1139,11 +1141,37 @@ static int sldns_wire2str_svcparam_ech2str(char** s,
...
+		} else {
+			w += sldns_str_print(s, slen, "%c", data[i]);
+		}
+	}
+
+	w += sldns_str_print(s, slen, "\"");
+
+	return w;
+}
+
```
wtoorop:
```suggestion
```
Escaped quotes and escaped backslashes are not printed correctly, but they can be in the input.
At least what you print should result into the same as what you read.
Together with the suggestion below, this can just be removed.

sldns/wire2str.c
```diff
@@ -1201,6 +1230,9 @@ int sldns_wire2str_svcparam_scan(uint8_t** d, size_t* dlen, char** s, size_t* sl
 	case SVCB_KEY_ECH:
 		r = sldns_wire2str_svcparam_ech2str(s, slen, data_len, *d);
 		break;
+	case SVCB_KEY_DOHPATH:
+		r = sldns_wire2str_svcparam_dohpath2str(s, slen, data_len, *d);
+		break;
```
wtoorop:
```suggestion
	case SVCB_KEY_DOHPATH:
		/* fallthrough */
```

sldns/str2wire.c
```diff
@@ -1511,7 +1516,70 @@ sldns_str2wire_svcbparam_alpn_value(const char* val,
...
+				expr_found++;
+			}
+		}
+
+		next_char = close_bracket+1;
+	}
+
+	if (expr_found != 1) {
+		return LDNS_WIREPARSE_ERR_SVCB_NO_DNS_VAR_IN_DOHPATH;
+	}
```
wtoorop: This scares me a bit because it does not anticipate dohpaths in utf-8. The curly bracket could be part of a utf-8 character. WDYT?

wtoorop (changes requested):

gthess (approved):
LGTM now and the other comments were addressed.

PR #757, opened 2022-09-20
Philip issue 749

Philip-NLnetLabs:
Based on issue 749, but now aligned with the OpenBSD version


gthess (approved):
LGTM

wcawijngaards (approved):
Yes, I think it is a good idea to align the source code.

PR #759, opened 2022-09-21
Add EDE (RFC8914) caching

### LINKED ISSUE #873: EDE information missing from reply to upstream
**Describe the bug**
Asked this question in the unbound mailing list (https://lists.nlnetlabs.nl/pipermail/unbound-users/2023-April/008069.html).
George (Yorgos) Thessalonikefs answered this is possible by using "EDE codes for DNSSEC validation failures to the SERVFAIL answers".

The FTL developer has added code to evaluate these codes, however, the results are inconsistent.

After investigating this, it turns out **the EDE codes are NOT always present in the reply**, the pi-hole topic has an attachment whith extended **log**ging from pihole-FTL (= dnsmasq) and a **pcap** file that clearly shows the missing EDE information. The pi-hole topic, start reading from here: https://discourse.pi-hole.net/t/dnssec-discussion-support-for-proxy-dnssec/62217/17

The **pcap** file and **log** is found in the next entry (18)

**To reproduce**
run unbound with "ede: yes", check the replies from unbound -> downstream.
- dig www.dnssec-failed.org (on linux host) -> OK
- browse (Microsoft Edge or Firefox) to www.dnssec-failed.org -> fail

**Expected behavior**
return EDE information for all queries, if available.
the strange thing is it works (EDE is returned) for dig requests (dig A www.dnssec-failed.org and dig AAAA www.dnssec-failed.org), but fails when a browser is making the request (open www.dnssec-failed.org in Microsoft edge and / or Firefox)

**System:**
 - Unbound version: 1.17.2 (compiled from GitHub source)
 - OS: raspbian on raspberry pi 3B (Linux raspberrypi 6.1.21-v7+ #1642 SMP Mon Apr  3 17:20:52 BST 2023 armv7l GNU/Linux)
 - `unbound -V` output:
```
Version 1.17.2
Configure line: --prefix=/usr --sysconfdir=/etc --disable-static --enable-tfo-client --enable-tfo-server --with-libevent --with-libhiredis --enable-cachedb --with-pidfile=/run/unbound.pid
Linked libs: libevent 2.1.12-stable (it uses epoll), OpenSSL 1.1.1n  15 Mar 2022
Linked modules: dns64 cachedb respip validator iterator
TCP Fastopen feature available
```
**Additional information**
The goal is to run pihole-FTL (= dnsmasq) with option proxy-dnssec. This appears to be working, but the EDE information isn't always present.

TCY16:


wcawijngaards (commented):
The code has todo items noted in the comments. The allocation of the string needs to be done in the region, when a reply is copied into a region. A reference to the malloced string is a problem when the original gets deleted, and also another new malloced string is a problem that would be leaked when the region is cleared.

gthess (commented):
Some initial comments:
- Needs subnetcache (ECS) and cachedb support;
- The long tests related to `unbound-control dump_cache` and `unbound-control load_cache` are currently failling.

daemon/cachedump.c
```diff
@@ -647,11 +651,15 @@ load_msg(RES* ssl, sldns_buffer* buf, struct worker* worker)
...
-	if(sscanf(s, " %u %u " ARG_LL "d %u %u %u %u", &flags, &qdcount, &ttl, 
-		&security, &an, &ns, &ar) != 7) {
+	if(sscanf(s, " %u %u " ARG_LL "d %u %u %u %u %d%n", &flags, &qdcount, &ttl,
+		&security, &an, &ns, &ar, &ede, &consumed) != 8) {
 		log_warn("error cannot parse numbers: %s", s);
 		return 0;
 	}
+	/* there may be EDE text after the numbers */
+	if(consumed > 0 && (size_t)consumed < strlen(s))
+		ede_str = s + consumed + 1 /* space */;
```
wcawijngaards: For both "" strings and NULL strings, there is an ede number a space and then end of line. The `consumed` is then one less then strlen, and `s+consumed+1` points to the end of string. This then later allocates a "" string for every item. But a NULL ptr is really wanted there.

Perhaps use `(size_t)consumed+1 < strlen(s)` as the guard on the if statement, that turns that situation into `ede_str` NULL and then NULL later on. Also "" strings are then turned into NULL, but there are few of those and it does not really change the EDE output, I believe.

daemon/worker.c
```diff
@@ -507,11 +507,11 @@ answer_norec_from_cache(struct worker* worker, struct query_info* qinfo,
 				msg->rep, LDNS_RCODE_SERVFAIL, edns, repinfo, worker->scratchpad,
 				worker->env.now_tv))
 					return 0;
-			/* TODO store the reason for the bogus reply in cache
-			 * and implement in here instead of the hardcoded EDE */
-			if (worker->env.cfg->ede) {
-				EDNS_OPT_LIST_APPEND_EDE(&edns->opt_list_out,
-					worker->scratchpad, LDNS_EDE_DNSSEC_BOGUS, "");
+			/* Attach the cached EDE (RFC8914) */
+			if(worker->env.cfg->ede) {
```
wcawijngaards: It now always attaches an EDE. It would be good to check that the ede code is not NULL or the string is not NULL, so that it does not attach an empty EDE. Perhaps modify edns_opt_list_append_ede to that it does not append an EDE with no content, or check in the if clause before calling it.

validator/val_sigcrypt.c
```diff
@@ -718,9 +718,9 @@ dnskey_verify_rrset(struct module_env* env, struct val_env* ve,
 	}
 	verbose(VERB_ALGO, "rrset failed to verify: all signatures are bogus");
 	if(!numchecked) {
-		*reason = "signature missing";
+		*reason = "signatures bogus";
```
wcawijngaards: I think this should still be signatures missing, and the EDE for MISSING RRSIGS below. But at line 731, after the next else clause, perhaps another else clause could be added. In that else clause, set the ede reason code to EDE_DNSSEC_BOGUS, and also the reason string can be set.

The numchecked refers to the number of signatures inspected, and if that is zero, there are no signatures that matched the keytag. Or no signatures are present.

validator/validator.c
```diff
@@ -2151,9 +2146,19 @@ processFinished(struct module_qstate* qstate, struct val_qstate* vq,
 				log_query_info(NO_VERBOSE, "validation failure",
 					&qstate->qinfo);
 			else {
-				char* err = errinf_to_str_bogus(qstate);
-				if(err) log_info("%s", err);
-				free(err);
+				char* err_str = errinf_to_str_bogus(qstate);
+				if(err_str) {
+					size_t err_str_len = strlen(err_str);
+					log_info("%s", err_str);
```
wcawijngaards, line 2152: The debug print out probably needs to be removed.
```suggestion
```

wcawijngaards (approved):
Code looks good. Caching errors could be very helpful for debugging attempts.

daemon/cachedump.c
gthess: I put space parsing in the format string since it is always there. Does the new code look okay?

daemon/worker.c
gthess: Added a check for the EDE code for this one; same as the other uses of edns_opt_list_append_ede.

validator/validator.c
gthess: I see that this log_info is also on master. Are you sure?

daemon/cachedump.c
wcawijngaards: Yes, the new code looks fine.

validator/validator.c
wcawijngaards: You are right, there is a check on the log level. It should not be removed.

validator/val_sigcrypt.c
gthess: I see your point but it is not 100% clear from the [rfc](https://www.rfc-editor.org/rfc/rfc8914.html#name-extended-dns-error-code-10-) if code 10 means there are no RRSIGS at all (we cover that case with `num==0` above) or no rrsigs that we are expecting. I interpreted the former, but I see that the later could also be the case. It mostly boils down to the end user looking at the EDE code, I believe it would be confusing to say rrsigs are missing when at least some rrsigs are there.
If you still think that missing rrsigs is the correct code there I have no problem changing this.

wcawijngaards: The ede code can be the bogus code. I was thinking the error text itself is not improving, when rrsigs that match certain conditions are absent. But, if the text is more germane for many cases, that would be fine for me.

gthess: Ah, if it is the text then I agree!

gthess:
Thanks @wcawijngaards for the review!
I believe the easiest way (because of all the changes here) to now include the feature for cachedb and ecs is to bring @TCY16's relevant commits here from #790.

PR #794, opened 2022-11-24
use a separate thread to write auth zones to disk

sakateka:
Hi!
We ran into a problem on an installation with very large auth-zones.
For example, we have auth-zone with 30k records and up to 70 MB in size.
At the time of the auth-zone update, they are written to disk in the same thread in which the event loop processing user requests is running.
This behavior greatly hurts the event loop, which causes user request timeouts.
For example, here is the moment of writing an auth-zone with a size of 70 MB, this operation can last up to 4 seconds.
```
bpftrace -e '
  uprobe:/usr/sbin/unbound-1.17.0-release:auth_zone_write_file {
    @start[tid] = nsecs;
  }
  uretprobe:/usr/sbin/unbound-1.17.0-release:auth_zone_write_file /@start[tid]/ {
    @ns[comm] = hist((nsecs - @start[tid]) / 1e6); delete(@start[tid]);
  }
'
Attaching 2 probes...
^C

@ns[unbound, 2020368]:
[64, 128)            4 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@                  |
[128, 256)           6 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[256, 512)           3 |@@@@@@@@@@@@@@@@@@@@@@@@@@                          |
[512, 1K)            0 |                                                    |
[1K, 2K)             0 |                                                    |
[2K, 4K)             2 |@@@@@@@@@@@@@@@@@                                   |
```
K - seconds

This means that this thread (when using SO_REUSEPORT) receives a user load, but cannot process it.
I made a simple patch that adds optional behavior for writing auth-zones in a separate thread.
Any suggestions are welcome!


sakateka:
This PR is intended to serve as a POC, although it can be modified for all supported OS.

PR #802, opened 2022-12-05
add validation EDEs to queries where the CD bit is set

TCY16:
at the request of @rijswijk :)


wcawijngaards (approved):
The code looks good to me. Does not look like the cache-response code path adds a EDE when CD is set, but perhaps that is not an issue for this code change.

PR #804, opened 2022-12-10
Fixed DNS Root Servers

madnuttah:
The DNS Root Servers have changed: https://www.iana.org/domains/root/servers

### DIFF
--- a/doc/example.conf.in
+++ b/doc/example.conf.in
@@ -1109,22 +1109,32 @@ remote-control:
 # sources of notifies.
 # auth-zone:
 #	name: "."
+#	primary: 198.41.0.4           # a.root-servers.net
 #	primary: 199.9.14.201         # b.root-servers.net
 #	primary: 192.33.4.12          # c.root-servers.net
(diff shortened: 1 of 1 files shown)


wcawijngaards (commented):
The root server addresses are correct. However, this is a list of root servers that allow zone transfer. And also the distribution servers are included in it. The list of all root servers can be set with configuration using a stub-zone option for zone "." and then a list of addresses. The builtin list of root servers is contained in a source file.

So the update tries to add all the root servers, but adding root servers that do not allow zone transfer is not good for the hyperlocal root zone configuration that tries to perform a zone transfer. Also removing the distribution servers is not helpful either. I think it is a case of mistaken list of servers.

wcawijngaards:
So, the patch is not a good idea to implement. The list of root servers is correct, but that list needs to be in another place. The root zone copy needs a list of servers that do zone transfer.

Thank you for checking the root server entries, as part of the review I also checked the builtin default root servers that unbound has against the list that the issue references. The builtin list is correct and complete at this time.

Issue #807, opened 2022-12-13
Install script fails due to space in filepath


sneurlax:
**Describe the bug**
When cross-compiling libunbound for x64 Windows from Ubuntu 20.04, [line 619 of Makefile.in](https://github.com/NLnetLabs/unbound/blob/master/Makefile.in#L619) causes `make install` to fail with error:

```
if test ! -e C:\Program Files\Unbound\service.conf; then /bin/bash ./install-sh -d `dirname C:\Program Files\Unbound\service.conf`; /bin/bash ./install-sh -c -m 644 doc/example.conf C:\Program Files\Unbound\service.conf; fi
/bin/bash: line 0: test: C:Program: binary operator expected
```

Because the string "Program Files" contains a space without being wrapped in quotes.

**To reproduce**
Build libunbound like:
```
#!/bin/bash

export SCRIPTDIR="$(pwd)"
export WORKDIR=${SCRIPTDIR}/build
export TYPES_OF_BUILD="x86_64"
export PREFIX=$WORKDIR/prefix_${TYPES_OF_BUILD}
export HOST=x86_64-w64-mingw32

UNBOUND_VERSION=release-1.17.0
UNBOUND_HASH="d25e0cd9b0545ff13120430c94326ceaf14b074f"
UNBOUND_SRC_DIR=$WORKDIR/unbound-1.17.0

cd $WORKDIR
rm -rf $UNBOUND_SRC_DIR
git clone https://github.com/NLnetLabs/unbound.git -b ${UNBOUND_VERSION} ${UNBOUND_SRC_DIR}
cd $UNBOUND_SRC_DIR
test `git rev-parse HEAD` = ${UNBOUND_HASH} || exit 1

./configure \
	CFLAGS=-fPIC \
	CXXFLAGS=-fPIC \
	--prefix=${PREFIX} \
	--host=${HOST} \
	--target=${HOST} \
	--enable-static \
	--disable-shared \
	--disable-flto \
	--with-pic \
	--with-ssl=${PREFIX} \
	--with-libexpat=${PREFIX}
make CROSS=x86_64-w64-mingw32.static- -j$THREADS
make install
```

**Expected behavior**
`make install` should not fail due to an error

**System:**
 - Unbound version: 1.17.0
 - OS: Unbunt 20.04
 - `unbound -V` output:
 ```
Command 'unbound' not found, but can be installed with:

sudo apt install unbound
```

**Additional information**
[...]

PR #808, opened 2022-12-13
Wrap Makefile script's directory variables in quotes

sneurlax:
This pull request wraps the directory variables on [line 619 of Makefile.in](https://github.com/NLnetLabs/unbound/blob/master/Makefile.in#L619) in quotes so as to avoid `make install` failure due to a script parsing error as seen in https://github.com/NLnetLabs/unbound/issues/807


sneurlax:
These changes have not been tested so as to not affect other platforms/toolchains; a reviewer ought to make sure that the main hosts/targets aren't compromised.

wcawijngaards (approved):
This change looks good to me. It should be fine for building on other targets.

wcawijngaards:
Thanks for the fix! Merged it into the code repository.

PR #826, opened 2023-01-13
Аdd a metric about the maximum number of collisions in lrushah

sakateka:
In continuation of the fix added here  90831af981221bbce1cd7b15055562336760e484
I would like to suggest adding a metric about the maximum number of collisions that happened in lruhash.
This will be the best indicator of a similar problem in the future.

### DIFF
--- a/daemon/remote.c
+++ b/daemon/remote.c
@@ -1065,6 +1065,11 @@ print_ext(RES* ssl, struct ub_stats_info* s, int inhibit_zero)
 		(unsigned)s->svr.infra_cache_count)) return 0;
 	if(!ssl_printf(ssl, "key.cache.count"SQ"%u\n",
 		(unsigned)s->svr.key_cache_count)) return 0;
+	/* max collisions */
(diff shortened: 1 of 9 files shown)


wcawijngaards (approved):
The change looks good. This code can be added.

wcawijngaards:
Thanks for the additional performance counter! Perhaps it can prevent other issues like this.

I have merged it in to the code repository. Also documentation is added, for the statistics entries, do you like the text for it; it can be seen in the commit https://github.com/NLnetLabs/unbound/commit/469133e8df8ecaf095d0d2fbe4f6b23f9ccebe23 ?

sakateka:
It looks very good, thank you!
I accidentally forgot about the documentation).

Perhaps it was worth writing `very large (hundreds)`
Because several hundred in one bin already affect performance.
In practice, during load testing, I have not seen a value greater than 20.

wcawijngaards:
Okay, improved the wording to add that detail! Useful to know what amount is trouble.

PR #827, opened 2023-01-13
Eliminate unnecessary Python reloading which causes memory leaks.

rcmcdonald91:
Reloading the Python interpreter for every HUP/reload is unnecessary. Python does not handle this very well and leaks memory.


rcmcdonald91:
This work also tipped me off to the lack of inplace callback cleanup in the python module. That has been corrected too

rcmcdonald91:
@wcawijngaards I hate to ping directly, but some visibility to this would be appreciated. Thanks :)

pythonmod/pythonmod.c
```diff
@@ -252,6 +253,16 @@ log_py_err(void)
...
 }
 
+/* we only want to unwind Python once at exit */
+void pythonmod_atexit(void)
+{
+   assert(py_mod_count == 0);
+   assert(maimthr != NULL);
+ 
+   PyEval_RestoreThread(mainthr);
+   Py_Finalize();
```
gthess: - Unbound prefers using `log_assert` instead of plain `assert`;
- This seems like a static function to me;
- Also small typo in `mainthr`.
```suggestion
static void
pythonmod_atexit(void)
{
   log_assert(py_mod_count == 0);
   log_assert(mainthr != NULL);
 
   PyEval_RestoreThread(mainthr);
   Py_Finalize();
```

pythonmod/pythonmod.c
```diff
@@ -310,6 +321,9 @@ int pythonmod_init(struct module_env* env, int id)
 #endif
       SWIG_init();
       mainthr = PyEval_SaveThread();
+
+      /* XXX: register callback to unwind Python at exit */
+      atexit(pythonmod_atexit);
```
gthess: Removing the XXX; I don't suppose there is anything more to do here.
```suggestion
      /* register callback to unwind Python at exit */
      atexit(pythonmod_atexit);
```

pythonmod/pythonmod.c
```diff
@@ -547,15 +561,15 @@ void pythonmod_deinit(struct module_env* env, int id)
...
-      }
+      py_mod_count--;
    }
    pe->fname = NULL;
    free(pe);
 
+   /* iterate over all possible callback types and clean up each in turn */
+   for (int cbtype = 0; cbtype < inplace_cb_types_total; cbtype++)
+      inplace_cb_delete(env, cbtype, id);
+
```
gthess: Good catch! It seems the callback delete functions are not exposed to Python.
Since the list of callbacks is really small (or non-existent) this works fine and keeps it simple for Python module users.
If we have more in the future we can revisit this part (well, together with the callback structures that would need to be something more than just a list).
But I would prefer `cbtype` to be declared at the start of the function like most Unbound code.
```suggestion
   /* iterate over all possible callback types and clean up each in turn */
   for(cbtype = 0; cbtype < inplace_cb_types_total; cbtype++)
      inplace_cb_delete(env, cbtype, id);

```

pythonmod/pythonmod.c
```diff
@@ -3,6 +3,7 @@
  *
  * Copyright (c) 2009, Zdenek Vasicek (vasicek AT fit.vutbr.cz)
  *                     Marek Vavrusa  (xvavru00 AT stud.fit.vutbr.cz)
+ * Copyright (c) 2023, Rubicon Communications, LLC (Netgate)
```
gthess: Not sure about the copyright note here for this small change.
Usually we don't do copyrights for fixes. If there is copyright in Unbound it is mainly from the original authors, as in this case.
Is the copyright addition something necessary in your case?
Btw acknowledgment for contributions are always noted in commit/merge messages, but also in the Changelog (edited by the Unbound development team) and release notes.

gthess (changes requested):
The changes look good, thanks!
I have some feedback and would need to do some local testing but this could be part of the next release.

pythonmod/pythonmod.c
rcmcdonald91: > * Unbound prefers using `log_assert` instead of plain `assert`;

Good to know. Thanks
> * This seems like a static function to me;

Yes. Thanks
> * Also small typo in `mainthr`.

Urgh. Thanks

pythonmod/pythonmod.c
rcmcdonald91: > Good catch! It seems the callback delete functions are not exposed to Python. Since the list of callbacks is really small (or non-existent) this works fine and keeps it simple for Python module users. If we have more in the future we can revisit this part (well, together with the callback structures that would need to be something more than just a list).

Agreed.

> But I would prefer `cbtype` to be declared at the start of the function like most Unbound code.

Done. Thanks

pythonmod/pythonmod.c
rcmcdonald91: > Not sure about the copyright note here for this small change. Usually we don't do copyrights for fixes. If there is copyright in Unbound it is mainly from the original authors, as in this case. Is the copyright addition something necessary in your case? Btw acknowledgment for contributions are always noted in commit/merge messages, but also in the Changelog (edited by the Unbound development team) and release notes.

Not necessary. Removed. Acknowledgement is fine in the commit/merge and Changelog. Thanks

pythonmod/pythonmod.c
rcmcdonald91: > Removing the XXX; I don't suppose there is anything more to do here.

Agreed. Thanks

gthess (approved):
LGTM with further testing! Valgrind at least locally stopped reporting the leaks.

PR #831, opened 2023-01-18
Improve Windows NSIS installer script (setup.nsi)

Pierre4012:
Two improvements of installer script :
1. avoid error message when Unbound is running,
2. add "DisplayVersion" in registry thus Windows package manager (Winget) can handle Unbound.

**1. Avoid error message :**

When Unbound is already installed and running, when the installer is used to update Unbund, it displays an error message :

error

In order to avoid this message, I suggest a modification to stop Unbound service if already installed.

**2. Adding "DisplayVersion" in registry :**

Winget use uninstall parameters in Windows registry to get program versions to be able to compare installed version and last available version to propose update. Without this parameter, Winget can't handle Unbound updates :

version

This parameter is also used to show program versions in the Windows remove program menu.

I suggest adding this parameter in the installer script.

### DIFF
--- a/winrc/setup.nsi
+++ b/winrc/setup.nsi
@@ -78,9 +78,17 @@ section "Root anchor - DNSSEC" SectionRootKey
 sectionEnd
 
 section "-hidden.postinstall"
-	# stop unbound service to allow file replacement
-	IfFileExists "$INSTDIR\unbound-service-remove.exe" 0 +2
-	nsExec::ExecToLog '"$INSTDIR\unbound-service-remove.exe" stop'
+	# if Unbund is already installed, ask to stop it to allow file replacement
+	IfFileExists "$INSTDIR\unbound-service-remove.exe" 0 next_label
+	MessageBox MB_YESNO "Should Unbound be stopped to permit the update ?" /SD IDYES IDNO false_label # defaults to yes on silent installations
+		nsExec::ExecToLog '"$INSTDIR\unbound-service-remove.exe" stop'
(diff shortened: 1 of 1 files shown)


gthess:
Hi, I like change no2 but not no1 :)
I would like for any termination of the already running service to be explicit and not because of the installer. Actually I would move `unbound.exe` up as to be the first file to be created, to avoid `LICENSE` and `README.txt` to be copied.

Pierre4012:
> Hi, I like change no2 but not no1 :) I would like for any termination of the already running service to be explicit and not because of the installer. Actually I would move `unbound.exe` up as to be the first file to be created, to avoid `LICENSE` and `README.txt` to be copied.

For no1 I can suggest adding a messagebox to ask user :

https://github.com/Pierre4012/unbound/commit/c66659f6675f53578a3f052032aa8ad5796af977

gthess:
I like it more, but maybe instead of Abort just Quit?
Now the installer window will just hang there with the aborted message and the only action is to "cancel" the installation?
Then maybe the message could be something like "Unbound is already installed! Would you like to stop the service to continue with the installation?"

Pierre4012:
Yes, simply Quit is much more adapted, have modified it.
I added your message proposal for MessageBox specifying "update".

https://github.com/Pierre4012/unbound/commit/a16aca17ff2d68efda2e1cb5d2213abf0b974ada

gthess (approved):
Thanks for this!

PR #840, opened 2023-02-06
unbound: fix no-cache feature

bzfbd:
Years ago I submitted the no-cache feature to unbound which over time broke. I debugged it down to the point that pre-seeding no_cache_store will fix this with minimal changes and make no-cache work again which remains highly valuable in a lot of environments this got deployed.

Obtained from:	https://reviews.freebsd.org/D38057

### DIFF
--- a/services/mesh.c
+++ b/services/mesh.c
@@ -63,6 +63,7 @@
 #include "util/data/dname.h"
 #include "respip/respip.h"
 #include "services/listen_dnsport.h"
+#include "iterator/iter_utils.h"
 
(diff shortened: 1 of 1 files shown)


gthess:
I suspect that this could be broken because of specific module configuration i.e., the `module-config:` option.
Do you have a specific configuration where this fails?
By default (`module-config: 'validator iterator'`) it works, as there is a testcase for that already.

bzfbd:
I believe last time I checked was with:
`module-config: "dns64 validator iterator`

PR #842, opened 2023-02-09
Ecs serve expired bug

Philip-NLnetLabs:


util/module.h
```diff
@@ -619,6 +619,12 @@ struct module_qstate {
 	/** if this is a validation recursion query that does not get
 	 * validation itself */
 	int is_valrec;
+#ifdef CLIENT_SUBNET
+	/** client network address is needed for the client-subnet option
+	 *  when probing, but we want use reply_list in mesh_info, because
+	 * we don't want to send a reply. */
```
gthess: Suggestion for more clear text
```suggestion
	/** client network address is needed for the client-subnet option
	 *  when prefetching, but we can't use reply_list in mesh_info, because
	 *  we don't want to send a reply for the internal query. */
```

doc/Changelog
```diff
@@ -1,3 +1,7 @@
+9 February 2023: Philip
+	- Fix #825: Unexpected behavior with client-subnet-always-forward
+	  and serve-expired
+
```
gthess: For PRs it is better to update the changelog after the merge, to avoid these conflicts and have the correct date.
(GitHub's suggestion does not seem to apply on all changed lines [1..4])
```suggestion
```

gthess (approved):
This looks good to me, some nits only.

Issue #860, opened 2023-03-09
Bad interaction with 0 TTL records and  serve-expired


saksham-m:
**Describe the bug**

NXDOMAIN will get stuck in the cache with serve-expired: yes when the new record is 0 TTL

**To reproduce**
1. serve-expired: yes
2. Let us say host.example.com is NXDOMAIN. This will be cached according to the SOA TTL 
3. Now add a new record for host.example.com with TTL 0.
4. Cache will forever serve NXDOMAIN as the new record is not allowed to be cached (Log shown: debug: TTL 0: dropped msg from cache)

**Expected behavior**
Cache should eventually recover without the need to flush the domain.

**Other notes**

1)

removing the 					FLAGS_GET_RCODE(rep->flags) != LDNS_RCODE_NXDOMAIN &&
check in line 629 of worker.c resolves the issue.

But of course, that means we will always resolve NXDOMAIN answers and not serve-expired

2) Setting cache-min-ttl: 1 also resolves the issue by forcing cache to write the new record.

**System:**
 - Unbound version:
 - OS: linux
 - `unbound -V` output:
 - 
 saksham@lhotse:/tmp/unbound-1.17.1$ sudo unbound -V
Version 1.17.1

Configure line:
Linked libs: mini-event internal (it uses select), OpenSSL 1.1.1f  31 Mar 2020
Linked modules: dns64 respip validator iterator

**Additional information**
Add any other information that you may have gathered about the issue here.

PR #866, opened 2023-03-22
0ttl

Philip-NLnetLabs:
Fix for issue #860

### DIFF
--- a/services/cache/dns.c
+++ b/services/cache/dns.c
@@ -178,7 +178,7 @@ dns_cache_store_msg(struct module_env* env, struct query_info* qinfo,
 		 * which could be useful for delegation information */
 		verbose(VERB_ALGO, "TTL 0: dropped msg from cache");
 		free(rep);
-		/* if the message is in then cache, remove that msg,
+		/* if the message is in the cache, remove that msg,
 		 * so that the TTL 0 response can be returned for future
 		 * responses (i.e. don't get answered from
 		 * cache, but instead go to recursion to get this TTL0
--- a/services/cache/dns.c
+++ b/services/cache/dns.c
@@ -134,24 +134,19 @@ msg_cache_remove(struct module_env* env, uint8_t* qname, size_t qnamelen,
 
 /** remove servfail msg cache entry */
 static void
-msg_del_servfail(struct module_env* env, struct query_info* qinfo,
+msg_del_for_0ttl(struct module_env* env, struct query_info* qinfo,
 	uint32_t flags)
 {
 	struct msgreply_entry* e;
-	/* see if the entry is servfail, and then remove it, so that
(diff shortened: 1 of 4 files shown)


services/cache/dns.c
```diff
@@ -183,12 +178,18 @@ dns_cache_store_msg(struct module_env* env, struct query_info* qinfo,
 		 * which could be useful for delegation information */
 		verbose(VERB_ALGO, "TTL 0: dropped msg from cache");
 		free(rep);
-		/* if the message is SERVFAIL in cache, remove that SERVFAIL,
+		/* if the message is in then cache, remove that msg,
```
wcawijngaards: spelling fix.
```suggestion
		/* if the message is in the cache, remove that msg,
```

testdata/serve_expired_0ttl_nodata.rpl
```diff
@@ -0,0 +1,154 @@
...
+
+stub-zone:
+	name: "example.com"
+	stub-addr: 1.2.3.4
+CONFIG_END
+
+SCENARIO_BEGIN Test serve-expired with NXDOMAIN followed by 0 TTL
+; Scenario overview:
+; - query for 0ttl.example.com. IN A
+; - answer from upstream is NODATA; will be cached for NORR_TTL(5)
```
wcawijngaards: Fix TTL value in comment.
```suggestion
; - answer from upstream is NODATA; will be cached for the SOA negative TTL.
```

testdata/serve_expired_0ttl_nxdomain.rpl
```diff
@@ -0,0 +1,154 @@
...
+
+stub-zone:
+	name: "example.com"
+	stub-addr: 1.2.3.4
+CONFIG_END
+
+SCENARIO_BEGIN Test serve-expired with NXDOMAIN followed by 0 TTL
+; Scenario overview:
+; - query for 0ttl.example.com. IN A
+; - answer from upstream is NXDOMAIN; will be cached for NORR_TTL(5)
```
wcawijngaards: Fix TTL value in comment.
```suggestion
; - answer from upstream is NXDOMAIN; will be cached for the SOA negative TTL.
```

wcawijngaards (approved):
I think it is a good idea to evict the message cache content when a 0 TTL new response is received. That synchronizes the content to the upstream content. The code looks good to have.

gthess (changes requested):
Looks good!
But I believe we don't need the `msg_del_for_0ttl()` function anymore, just straight `msg_cache_remove()` in that place.
I would also like to see the same behavior for the ECS cache and cachedb.

Philip-NLnetLabs:
Yes, I'll change it to call msg_cache_remove. Maybe you can to the other two issues in a separate ticket.

PR #867, opened 2023-03-23
Unbound host tls

Philip-NLnetLabs:
Fix for unbound-host (libunbound) when tls is requested without explicit certs

### DIFF
--- a/libunbound/libworker.c
+++ b/libunbound/libworker.c
@@ -168,14 +168,11 @@ libworker_setup(struct ub_ctx* ctx, int is_bg, struct ub_event_base* eb)
 		hints_delete(w->env->hints);
 		w->env->hints = NULL;
 	}
-	if(cfg->ssl_upstream || (cfg->tls_cert_bundle && cfg->tls_cert_bundle[0]) || cfg->tls_win_cert) {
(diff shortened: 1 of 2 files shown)


libunbound/libworker.c
```diff
@@ -168,14 +168,11 @@ libworker_setup(struct ub_ctx* ctx, int is_bg, struct ub_event_base* eb)
...
-		if(!w->sslctx) {
-			/* to make the setup fail after unlock */
-			hints_delete(w->env->hints);
-			w->env->hints = NULL;
-		}
+	w->sslctx = connect_sslctx_create(NULL, NULL,
+		cfg->tls_cert_bundle, cfg->tls_win_cert);
+	if(!w->sslctx) {
+		libworker_delete(w);
+		return NULL;
```
wcawijngaards, line 175: This return failure does not unlock the lock cfglock that happens on line L181. The solution that used to be used is suggested in the code block.
```suggestion
		/* to make the setup fail after unlock */
￼		hints_delete(w->env->hints);
￼		w->env->hints = NULL;
```

wcawijngaards (approved):
This looks okay. The cfglock needs to be unlocked on the allocation failed code path for the SSL context, if it was locked.

PR #880, opened 2023-04-24
services/authzone.c: remove redundant check

chipitsine:
found by cppcheck

services\authzone.c:7513:12: style: Condition 'rrlist[i]' is always true [knownConditionTrueFalse]


gthess:
Thanks!

PR #882, opened 2023-04-26
Features/dropqueuedpackets

vvfedorenko:
This pull request adds `sock_queue_timeout` configuration options which stores the amount of seconds spend in the socket queue after which the packet is considered old and should be dropped. The reason to drop is because client won't be waiting for the reply anymore and there is no reason to process the query.
There are also 2 metrics added:
 - `num.queries_timed_out` to store the amount of dropped packets
 - `query.queue_time_us.max` to collect the max time that was found for packets to wait in the socket queue (this can be easily converted to histogram if needed)


util/config_file.c
```diff
@@ -543,6 +544,7 @@ int config_set_option(struct config_file* cfg, const char* opt,
 	else S_NUMBER_NONZERO("tcp-reuse-timeout:", tcp_reuse_timeout)
 	else S_YNO("edns-tcp-keepalive:", do_tcp_keepalive)
 	else S_NUMBER_NONZERO("edns-tcp-keepalive-timeout:", tcp_keepalive_timeout)
+	else S_NUMBER_NONZERO("sock-queue-timeout:", sock_queue_timeout)
```
wcawijngaards: With this change it matches the conf parser, where 0 is accepted.
```suggestion
	else S_NUMBER_OR_ZERO("sock-queue-timeout:", sock_queue_timeout)
```

util/netevent.c
```diff
@@ -833,6 +846,8 @@ comm_point_udp_ancil_callback(int fd, short event, void* arg)
 #ifndef S_SPLINT_S
 	struct cmsghdr* cmsg;
 #endif /* S_SPLINT_S */
+	struct timespec *ts;
+	struct timeval *tv;
```
wcawijngaards: The compiler says this is an unused variable.
```suggestion
```

util/netevent.c
```diff
@@ -894,8 +910,22 @@ comm_point_udp_ancil_callback(int fd, short event, void* arg)
...
+				cmsg->cmsg_type == SO_TIMESTAMPNS) {
+				ts = (struct timespec *)CMSG_DATA(cmsg);
+				TIMESPEC_TO_TIMEVAL(&rep.c->recv_tv, ts);
+			} else if( cmsg->cmsg_level == SOL_SOCKET &&
+				cmsg->cmsg_type == SO_TIMESTAMPING) {
+				ts = (struct timespec *)CMSG_DATA(cmsg);
+				TIMESPEC_TO_TIMEVAL(&rep.c->recv_tv, ts);
+			} else if( cmsg->cmsg_level == SOL_SOCKET &&
+				cmsg->cmsg_type == SO_TIMESTAMP) {
+				rep.c->recv_tv = *(struct timeval *)CMSG_DATA(cmsg);
```
wcawijngaards: Even though I do not know of platforms with both timestamping and alignment issues, the code change is to be careful with unaligned memory reads.
```suggestion
				memmove(&rep.c->recv_tv, (struct timeval *)CMSG_DATA(cmsg), sizeof(struct timeval));
```

wcawijngaards (approved):
Thank you for the nice code and some code cleanup, that implements the timestamping feature! The code looks nice, I have but a few smaller suggestions.

Is it okay if I put the banner over the new files util/timeval_func.c and util/timeval_func.h, that contain the moved timeval functions from other places in the source code? I mean to put the license text on the file, like we have for other files. And the Copyright statement, for NLnet Labs above the file as well, like for other files. That makes it similar to the other code we have. NLnet Labs wants to have the copyright and also the license on the files to maintain the project. I think it is nice to clean up the code with them.

util/netevent.c
vvfedorenko: Ah, yeah, it's an artefact of the previous version, we can safely remove it.

util/netevent.c
vvfedorenko: Even though I'm more about to rely on compiler in such cases, I'm ok with this change. But there is no need for cast anymore with memmove.

util/config_file.c
vvfedorenko: Ah, yeah, you are right, missed this part, thanks!

vvfedorenko:
Yep, sure, you can put statements, there is no new code in these files, it's just copy-paste and cleanup.
Let me rebase and update the PR with the suggestions applied

util/netevent.c
wcawijngaards: Yes, that is true. I'll mark this as resolved then.

wcawijngaards:
With the adjustments, I'll approve and merge this. Thank you for the additional feature and cleanup!

After that, I will likely make a couple of smaller changes; spelling error, documentation mentions, and Changelog entry, to adjust the code a little. Thanks for the header license change, that is what I wanted.

PR #884, opened 2023-04-29
Implement block_aaaa similar to block_a

strlcat:
Reason behind this is that certain ancient apps
get confused with IPv6 and stuff, so offer possibility to make them work too.

### DIFF
--- a/doc/example.conf.in
+++ b/doc/example.conf.in
@@ -820,6 +820,8 @@ server:
 	#   that name
 	# o block_a resolves all records normally but returns
 	#   NODATA for A queries and ignores local data for that name
+	# o block_aaaa resolves all records normally but returns
+	#   NODATA for AAAA queries and ignores local data for that name
 	# o always_null returns 0.0.0.0 or ::0 for any name in the zone.
 	# o noview breaks out of that view towards global local-zones.
 	#
--- a/doc/unbound.conf.5.in
+++ b/doc/unbound.conf.5.in
@@ -1406,7 +1406,7 @@ address space are not validated.  This is usually required whenever
 Configure a local zone. The type determines the answer to give if
 there is no match from local\-data. The types are deny, refuse, static,
 transparent, redirect, nodefault, typetransparent, inform, inform_deny,
-inform_redirect, always_transparent, block_a, always_refuse, always_nxdomain,
+inform_redirect, always_transparent, block_a, block_aaaa, always_refuse, always_nxdomain,
 always_null, noview, and are explained below. After that the default settings
 are listed. Use local\-data: to enter data into the local zone. Answers for
 local zones are authoritative DNS answers. By default the zones are class IN.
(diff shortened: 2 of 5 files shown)


winooxx:
nice implemetation

jailbird777:
Any change to get this merged?

strlcat:
I actually made a good use of this for sites and domains that are too restrictive and would ban outright public IPv6 tunnel brokers for example (as I "sit" on one of them) together with netfilter filtering aswell.
Would be a nice feature to have...

gauthamw-stripe:
Gentle ping :)

Could we look to merge this? This is quite similar to [the block_a zone-type implementation](https://github.com/NLnetLabs/unbound/commit/d5b9a790fef446ced08ee84023f6bcb6c31c1a95) and would help our network a fair bit too!

rozhuk-im:
@wcawijngaards Please!

wcawijngaards:
The item belongs to @Philip-NLnetLabs . And he tells me that he does not want it because it breaks DNSSEC. DNSSEC would not work for downstream validators. The code looks nice otherwise.

rozhuk-im:
But it do same as ```block_a```, may be just add note to docs about this?

gauthamw-stripe:
I've not read how this breaks DNSSEC but the fix for this should be identical for both `block_a` (already merged) and `blocked_aaaa`. There are systems that don't use DNSSEC today (and might not anytime soon, eg: private-only networks) that could benefit from this so can we move the DNSSEC support into another issue?

ashwin316:
Hello 
Are we looking forward to merge this change? This looks fine to me also.
This will help our internal network in reducing a good amount of network calls :)

mmiller7:
Ended up here because "private-address: ::/0" seems to be breaking some clients as it seems to return SERVFAIL adding that line...and i want no IPv6 addresses because my ISP doesn't do IPv6 and so clients take ages timing out on IPv6 before they use IPv4

We REALLY need some way to block IPv6 addresses in DNS when you know the ISP doesn't support DNS so that clients don't lag trying to connect to something it will never successfully reach.

gthess:
@mmiller7 , if you need something global (and not per domain functionality that this issue is about) you can use:
```
server:
    ...
    module-config: "respip validator iterator" # add your other modules as well but respip is needed
    response-ip: ::/0 redirect
```

mmiller7:
> @mmiller7 , if you need something global (and not per domain functionality that this issue is about) you can use:
> 
> ```
> server:
>     ...
>     module-config: "respip validator iterator" # add your other modules as well but respip is needed
>     response-ip: ::/0 redirect
> ```

Oh cool, I haven't found that anywhere else...did seem to do the trick!

gauthamw-stripe:
@Philip-NLnetLabs - sorry for the repeated pings but what do you think about this proposal? This is identical to an existing feature `block_a` so hopefully should not increase maintenance cost much.

If there's a concern about this working with DNSSEC, I think this is already broken with `block_a` and we can separately track a fix for both of them, right? This feature does provide value-add so it'd be great if we could merge this. Adding on, DNSSEC is not widely used especially in internal cloud environments so there's probably a fairly large user-base that wouldn't be concerned by the lack of compatibility b/w this feature and DNSSEC.

Philip-NLnetLabs:
Why not rework the patch to fix the DNSSEC issues first?

gauthamw-stripe:
Thanks for the response, Philip! Is there any documentation around the breakage for DNSSEC? Apart from the comment above by Wouter, I don't think this has been documented anywhere.

If the fix is straight-forward, maybe we can fix it as part of this PR. If this is a complex fix, I'd suggest again that we track it separately since noone has raised concerns with `block_a` and fixing for `block_aaaa` won't increase that effort. Note that I am not the original author of this PR or the original one for `block_a` and have no experience with DNSSEC 😓 

What do you think?

Philip-NLnetLabs:
For DNSSEC, there are two things that need to be done, preferably with additional tests to show that it works:
1) if an address is deleted, make sure that the reply does not have the AD bit set.
2) If the request has to DO bit set, then do not change anything, just return the unmodified reply.

PR #892, opened 2023-05-31
Add cachedb hit stat

gthess:


daemon/stats.c
```diff
@@ -356,6 +356,11 @@ server_stats_compile(struct worker* worker, struct ub_stats_info* s, int reset)
 	s->svr.num_query_subnet = 0;
 	s->svr.num_query_subnet_cache = 0;
 #endif
+#ifdef USE_CACHEDB
+	s->svr.num_query_cachedb += (long long)worker->env.mesh->ans_cachedb;
```
wcawijngaards: Should this be without the addition, like the other initialisations in this routine? Like the initialisation of mesh_jostled = ..mesh->stats_jostled; and others.

```suggestion
	s->svr.num_query_cachedb = (long long)worker->env.mesh->ans_cachedb;
```

testdata/stat_values.tdir/stat_values.test
```diff
@@ -414,4 +414,98 @@ rrset.cache.count=3
 infra.cache.count=2"
 
 
+if test x$USE_CACHEDB == "x1"; then
```
wcawijngaards: The shell test command uses '=' for string comparison. Perhaps this change is better.

```suggestion
if test x$USE_CACHEDB = "x1"; then
```

wcawijngaards (approved):
Code looks good, check the `+=` vs `=` for initialisation in stats.c.

daemon/stats.c
gthess: Not sure anymore :) I was following the same logic as with `ans_expired` above.
If that is correct maybe moving this up there would be better for readability.

wcawijngaards: The addition is used for `ans_secure` and others, where the value is incremented through `worker->stats` and also has a counter in the mesh. These are then added in. The `ans_cachedb` counter is only incremented through the counter referenced, so addition is not really necessary. But it is added to a zero value in the `worker->stats` structure that was copied some lines earlier, so the change is not really necessary.

gthess: Ah right, now I see it! I was confused because of the same name in worker stats and mesh stats.

gthess:
Thanks for the review @wcawijngaards!
I am tagging this for the next release (pending user feedback for merging if it arrives in a timely manner)

Issue #887, opened 2023-05-08
Timeouts to forward servers on BSD based system with ASLR


Jakker:
Reported on [FreeBSD bugzilla](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270912) that on BSD systems with ASLR enabled that there are unexpected timeouts using forwarders. One wonders why this is not happening (or has not been noticed) on Linux type systems.

There is not enough information (yet) to judge what causes the problem.

montykabinski:
After an upgrade to FreeBSD 14.2-RELEASE this bug may cause seriously unreliable operation. The operator of the system will have no warning or any obvious way to understand why a formerly solid Unbound installation has suddenly become flaky.

elfctl -e +noaslr /usr/local/sbin/unbound
does work around the problem, but is an obscure solution for a difficult-to-diagnose problem.

grahamperrin:
> 14.2-RELEASE

1*3*.2-RELEASE, yes?

(We have not yet reached release of 14.0)

----

Also, for reference: [elfctl(1)](https://man.freebsd.org/cgi/man.cgi?query=elfctl&sektion=1&manpath=freebsd-release)

montykabinski:
Yes sorry, fat-fingered it, meant to type 13.2-RELEASE

rcmcdonald91:
We are also seeing this in pfSense as well. Let me know how I might be of assistance here.

One thing that would be interesting to test is if this is unique to a specific version of Unbound. Has anyone tried replicating this with an older version of Unbound?

Has anyone that can replicate this tried on the latest 14-CURRENT?

Reproducing this internally has been difficult so it would swell if those that can reproduce it reliably test these additional cases so we can start narrowing this down.

mzary:
> After an upgrade to FreeBSD 14.2-RELEASE this bug may cause seriously unreliable operation. The operator of the system will have no warning or any obvious way to understand why a formerly solid Unbound installation has suddenly become flaky.

Please don't scare people. This issue is probably configuration specific. I am running a bunch of dns/unbound and never have seen a problem manifesting this way. The PR on FreeBSD Bugzilla mentions forward-tls-upstream: yes option as triggering this issue.

Maltz42:
[...]

Issue #516, opened 2021-08-01
Stream reuse does not work on Windows (rbtree issue?)


yan12125:
**Describe the bug**

I use `forward-addr: 1.0.0.1@853#cloudflare-dns.com` as the DoT upstream. On Windows, connections are apparently not reused as there are many TCP connections, and the latency for the first query after restarting unbound is much higher than Linux and Android devices in the same local network.

**To reproduce**

1. Run `unbound.exe -ddpvvvvv -c unbound.conf 2>&1 > unbound.log` from PowerShell. The content of unbound.conf is:
```
server:
    verbosity: 1
    tls-win-cert: yes
    auto-trust-anchor-file: "C:\Program Files\Unbound\root.key"

forward-zone:
    name: "."
    forward-tls-upstream: yes
    forward-addr: 1.0.0.1@853#cloudflare-dns.com
```
2. Run `nslookup example.com`

**Expected behavior**

`netstat -ano` reports only one connection to 1.0.0.1:853, and the nslookup command returns the result in a short time (e.g., <200ms)

**System:**
 - Unbound version: tested both 1.13.1 from the official installer and 1.13.2 (commit b6abcb150808412678e87e49c829b1a3ef6509da) I built locally
 - OS: Windows 10 20H2
 - `unbound -V` output:

```
Version 1.13.2

Configure line: --host=x86_64-w64-mingw32 --target=x86_64-w64-mingw32 --build= --prefix=/usr/x86_64-w64-mingw32 --libdir=/usr/x86_64-w64-mingw32/lib --includedir=/usr/x86_64-w64-mingw32/include --enable-shared --enable-static --enable-static-exe
Linked libs: event winsock (it uses WSAWaitForMultipleEvents), OpenSSL 1.1.1k  25 Mar 2021
Linked modules: dns64 respip validator iterator

BSD licensed, see LICENSE in source package for details.
Report bugs to unbound-bugs@nlnetlabs.nl or https://github.com/NLnetLabs/unbound/issues
```

```
Version 1.13.1

[...]

PR #894, opened 2023-06-09
Fix for issue #887 (Timeouts to forward servers on BSD based system w…

Philip-NLnetLabs:
…ith ASLR)

and proabbly #516 (Stream reuse does not work on Windows)


wcawijngaards (approved):
These fixes look sensible. They are both checks for a return of the element that is just before, and thus smaller than, the key value that is looked for. Good to fix the code for it!

PR #896, opened 2023-06-11
Fix: #895: pythonmodule: add all site-packages directories to sys.path

### LINKED ISSUE #895: python + sysconfig gives ANOTHER path comparing to distutils
```
$ python3 
Python 3.11.3 (main, May 24 2023, 00:00:00) [GCC 13.1.1 20230511 (Red Hat 13.1.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import distutils.sysconfig
>>> print(distutils.sysconfig.get_python_lib(1,0))
/usr/lib64/python3.11/site-packages
>>> import sysconfig
>>> print(sysconfig.get_path("platlib"))
/usr/local/lib64/python3.11/site-packages
```

So, the changes around #769 make AGAIN

`error: pythonmod: cannot initialize core module: unboundmodule.py`

Is it possible not to hardcode (at ./configure stage and then in Makefile) the path at all ?

https://bugzilla.redhat.com/show_bug.cgi?id=2213996
https://github.com/mesonbuild/meson/issues/10459#issuecomment-1147462637

socketpair:
1. Actually, adding all site-packages dir is good. since pythonmodule should be able to import other Python libraries.
2. everything should work even since Python 2.6, no `#ifdef` required.

### DIFF
--- a/pythonmod/pythonmod.c
+++ b/pythonmod/pythonmod.c
@@ -328,44 +328,12 @@ int pythonmod_init(struct module_env* env, int id)
          env->cfg->directory);
          PyRun_SimpleString(wdir);
       }
-      /* Check if sysconfig is there and use that instead of distutils;
-       * distutils.sysconfig is deprecated in Python 3.10. */
-#if PY_MAJOR_VERSION <= 2 || (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION <= 9)
-      /* For older versions, first try distutils.sysconfig because the
-       * sysconfig paths may contain wrong values, eg. on Debian10 for
-       * python 2.7 and 3.7. */
-      if(PyRun_SimpleString("import distutils.sysconfig \n") < 0) {
-         log_info("pythonmod: module distutils.sysconfig not available; "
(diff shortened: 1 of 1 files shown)


socketpair:
`site.getsitepackages()` INCLUDE all dirs specified in `distutils.sysconfig.get_python_lib(1,0)` and in `sysconfig.get_path('platlib')`

wcawijngaards (approved):
The code change seems to be both simpler and better at picking up a list of python site-packages paths. There is also a site.getusersitepackages() directory, but that used to not be added previously. The site module seems to work for python 2.7 and python 3.6 and 3.10.

socketpair:
@wcawijngaards what should be done in order to merge ?

wcawijngaards:
Waiting for ci, it has finished, so it looks fine to merge.

wcawijngaards:
Thanks for the code contribution for the python integration! Hopefully this works to detect the python plugin install location to make the pythonmod run its python script in unbound and process DNS queries.

PR #901, opened 2023-06-19
config: improve handling of unknown modules

trofi:
The change fixes module print when specified module is unknown. On example config:

    server:
      module-config: "respip valdator iterator"

Before the change printed error looked like:

    error: Unknown value in module-config, module: ''. This module is
    not present (not compiled in), See the list of linked modules with
    unbound -V

After the change module is printed as expected:

    error: Unknown value in module-config, module: 'valdator'. This
    module is not present (not compiled in), See the list of linked
    modules with unbound -V

Module truncation happens because parse error does not guarantee that leading whitespace is removed by `module_factory()` call.

The change always removes leading whitespace (if present).


gthess:
Thanks!

PR #902, opened 2023-06-20
DNS Error Reporting (RFC 9567)

wtoorop:
Draft PR for work on ~[draft-ietf-dnsop-dns-error-reporting](https://datatracker.ietf.org/doc/draft-arends-dns-error-reporting/)~ [RFC 9567](https://datatracker.ietf.org/doc/rfc9567/)


gthess:
I believe this is now ready for review.
There are no limits for error reporting.
Reports will be generated for client inited queries that gave an error during the mesh state.
This is mostly the validator at the moment. When EDEs for RPZ are introduced (blocked, censored, etc) it needs to be revisited (probably by excluding those codes from the report).
LDNS_EDE_OTHER is excluded from the reports because the code does not make sense by itself.
LDNS_EDE_STALE_ANSWER is not interfering with the report (it is generated when responding with stale data, not the same code path).

util/config_file.c
```diff
@@ -410,6 +410,8 @@ config_create(void)
 	cfg->ipset_name_v6 = NULL;
 #endif
 	cfg->ede = 0;
+	cfg->ede_serve_expired = 0;
```
wcawijngaards, line 420: If this line is moved to here, then the line that sets ede_serve_expired at line 287 should be removed.

wcawijngaards (approved):
The code looks fine to include. The option set default off is a safe way to introduce the feature.

gthess:
Thanks @wcawijngaards ! And nice catch!

gthess:
@wcawijngaards, I merged and added a new stats item for dns_error_reports. Do you want to have a final look for completeness?

wcawijngaards:
The changes look fine! Good to have stats for the new feature.

PR #903, opened 2023-06-22
contrib: add yocto compatible init script

beni-sandu:
This has been used on yocto side for quite a while: http://cgit.openembedded.org/meta-openembedded/tree/meta-networking/recipes-support/unbound/unbound/0001-contrib-add-yocto-compatible-init-script.patch

Let's add a separate yocto compatible init script that we can maintain, allowing us to drop the patch from openembedded side.

Looks like I cannot add reviewers myself, and I don't know who should review this, I will cc some folks below:

@wcawijngaards @gthess 

Thanks.

### DIFF
--- a/contrib/unbound.init_yocto
+++ b/contrib/unbound.init_yocto
@@ -1,0 +1,139 @@
+#!/bin/sh
+#
+# unbound	This shell script takes care of starting and stopping
+#		unbound (DNS server).
+#
+# chkconfig:   - 14 86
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
This init file looks fine to use on a Linux system. Thank you for the contribution!

wcawijngaards:
Thank you for the addition! I have put a mention in the contrib/README file and merged it into the repository.

PR #907, opened 2023-07-07
Subnet cache prefetch

gthess:
This is #664 with some changes and re-merging for current master.


wcawijngaards (approved):
The easier code flow commit looks nice.

PR #911, opened 2023-07-12
Exclude EDE before other EDNS options when there isn't enough space

natalie-reece:
This change drops any EDE options from a response packet when there is not enough space for them but there is still enough space for the rest of the OPT record. This matches with the behavior specified in RFC 8914, which says "When the response grows beyond the requestor's UDP payload size, servers SHOULD truncate messages by dropping EDE options before dropping other data from packets".

In the function reply_info_answer_encode, I made the following changes:
- Instead of reserving space for the entire edns record, space is reserved for the size the record would be without any ede options.
- After the rest of the reply is added, if there is enough space for the entire edns record, it is added to the response.
- Otherwise, if there would be enough space without ede options, they are removed and the edns record is added without them.

In order to do this, I added a function calc_edns_option_size that calculates the size of an edns option. This function is similar to the existing function calc_edns_field_size, except it only finds the size for one option code.

The function error_encode was also changed so that if there is not enough space for all of the edns data, then any ede options are removed. Then, if there is enough space for the rest of the edns data, it is added, otherwise it is left out.

### DIFF
--- a/util/data/msgencode.c
+++ b/util/data/msgencode.c
@@ -806,6 +806,24 @@ calc_edns_field_size(struct edns_data* edns)
 	return 1 + 2 + 2 + 4 + 2 + rdatalen;
 }
 
+uint16_t
+calc_edns_option_size(struct edns_data* edns, uint16_t code)
+{
+	size_t rdatalen = 0;
+	struct edns_option* opt;
+	if(!edns || !edns->edns_present)
+		return 0;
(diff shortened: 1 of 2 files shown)


gthess:
Thanks for this!
I went a step further and tried to see if EDEs (excluding OTHER) could be attached without their EXTRA-TEXT field in case the code alone could prove useful for debugging.

PR #930, opened 2023-09-05
add void to log_ident_revert_to_default declaration

sthen:
Avoid warning from LLVM 16:

util_log.c:190:33: warning: a function declaration without a prototype is deprecated in all versions of C              [-Wstrict-prototypes]                                                                                                  void log_ident_revert_to_default()                                                                                    
                                ^                                                                                     
                                 void


gthess:
Thanks!

PR #955, opened 2023-10-17
fix ipset wrong behavior

buevsan:
Issue description:

    If local zone directive is assigned as following:

      "local-zone domain.com ipset"

    then any answers for query with names such as
    "anytext1domain.com" or "example2domain.com"
    will be added to ipset.

  This commit fixes this issue.


gthess (approved):
Looks good, thanks!

PR #968, opened 2023-11-24
Replace the obsolescent fgrep with grep -F in tests

gthess:
On newer systems I get the message:
```
fgrep: warning: fgrep is obsolescent; using grep -F
```
Looking around I see that `grep -F` is universally supported
- https://man.freebsd.org/cgi/man.cgi?query=grep&sektion=1
- https://man.openbsd.org/grep
- https://man.netbsd.org/grep.1
- https://www.man7.org/linux/man-pages/man1/grep.1.html
- https://ss64.com/osx/grep.html
- https://docs.oracle.com/cd/E88353_01/html/E37839/grep-1.html


wcawijngaards (approved):
The changes look fine. I could find no problems with it.

gthess:
Thanks!

PR #973, opened 2023-12-06
Use the origin (DNAME) TTL for synthesized CNAMEs as per RFC 6672

gthess:


wcawijngaards (approved):
The code looks good and makes the CNAME TTL for synthesized CNAMEs that come after a DNAME record equal to the DNAME TTL, so not 0. That improves caching so that could be useful, in the few cases where DNAMEs are used in practice.

gthess:
Thanks @wcawijngaards!

PR #975, opened 2023-12-07
Fixed some syntax errors in rpl files.

Philip-NLnetLabs:


wcawijngaards (commented):
Approved, these are good syntax fixes for a couple of rpl files.

Issue #970, opened 2023-11-30
DNS response to negative cache ttl is not rfc compliant


ShipeiXu:
**Describe the bug**
RFC 2308 stipulates that for authoritative server responses to nxdomain, the parser unbound needs to be cached according to the minimum of ttl and minimum ttl. But I observed that unbound did not adopt this standard, but directly used the cache-max-negative-ttl value. This is not expected.

"Name servers authoritative for a zone MUST include the SOA record of the zone in the authority section of the response when reporting an NXDOMAIN or indicating that no data of the requested type exists. This is required so that the response may be cached. "

**To reproduce**
1. dig @127.0.0.1 mail.cinda.com.cn
2. result is: 
  cinda.com.cn.		597	IN	SOA	cinda.com.cn. root.cinda.com.cn. 2006080401 1 86400 2419200 1

**Expected behavior**
the result need to be:
cinda.com.cn.		1	IN	SOA	cinda.com.cn. root.cinda.com.cn. 2006080401 1 86400 2419200 1

**System:**
 - Unbound version: Version 1.14.0
 - OS: centos 3.10.0-1160.80.1.el7.x86_64
 - `unbound -V` output:
Version 1.14.0
Configure line: --with-libnghttp2 --prefix=/usr/unbound/ --enable-subnet --with-pthreads --with-libevent --enable-dnstap --enable-cachedb --enable-ipsecmod --enable-ipset --enable-linux-ip-local-port-range --enable-dnscrypt --enable-systemd --with-pythonmodule
Linked libs: libevent 2.0.21-stable (it uses epoll), OpenSSL 1.0.2k-fips  26 Jan 2017
Linked modules: dns64 python cachedb ipsecmod subnetcache ipset respip validator iterator
DNSCrypt feature available

BSD licensed, see LICENSE in source package for details.
Report bugs to unbound-bugs@nlnetlabs.nl or https://github.com/NLnetLabs/unbound/issues

**Additional information**
Add any other information that you may have gathered about the issue here.

wcawijngaards:
When I run this query, I get the result with 1 TTL for the SOA. The changelog says this was fixed in 1.13.1, so it should work in 1.14.0. It seems to work with the latest version, from the code repository. It does not seem to go wrong, it correctly picks up the minimum of the SOA TTL and the SOA MINIMUM TTL. And the answer is returned with a SOA record with a TTL of 1 for the query from the issue.

ShipeiXu:
“To reproduce:
[...]

PR #978, opened 2023-12-12
fix: the conflict of cache-min-ttl and cache-max-negative-ttl

ShipeiXu:
cache-min-ttl/cache-max-ttl与cache-max-negative-ttl相互独立。

### DIFF
--- a/util/data/msgreply.c
+++ b/util/data/msgreply.c
@@ -225,17 +225,15 @@ rdata_copy(sldns_buffer* pkt, struct packed_rrset_data* data, uint8_t* to,
 		 * minimum-ttl in the rdata of the SOA record */
 		if(*rr_ttl > soa_find_minttl(rr))
 			*rr_ttl = soa_find_minttl(rr);
-	}
-	if(!SERVE_ORIGINAL_TTL && (*rr_ttl < MIN_TTL))
-		*rr_ttl = MIN_TTL;
-	if(!SERVE_ORIGINAL_TTL && (*rr_ttl > MAX_TTL))
-		*rr_ttl = MAX_TTL;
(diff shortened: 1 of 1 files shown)


gthess:
Hi @ShipeiXu, I believe the current logic is correct:
1. Check if the minimum SOA TTL needs to be used;
1. Constrain the TTL value between the global cache-min-ttl and cache-max-ttl configuration values if any;
1. Make sure cache-max-negative-ttl is honored for negative answers.

With the proposed changes, the global cache-min-ttl and cache-max-ttl options would not apply to negative answers.

With the current configuration options (and I assume you have configured cache-min-ttl already, thus your issue) do I understand correctly that you would like to instruct the cache to not cache for lower values than cache-min-ttl but you would like an exception for negative answers?

ShipeiXu:
You can do whatever you want without considering RFC 2308. However, RFC 2308 requires that negative caching be cached according to the minimum ttl of the authorization zone in the authoritative response. Therefore cache-max-negative-ttl should not be affected by other parameters. If you need to limit the negative cache minimum, you should consider adding a cache-min-negative-ttl instead of using cache-min-ttl. Negating the cached TTL will cause a lot of confusion for authoritative managers to add and delete records.

ShipeiXu:
Your understanding is correct, I did set cache-min-ttl. But for negative caching, this is not my expected result. Regardless of whether you modify it or not, I will use my logic in my own version

gthess:
Hi @ShipeiXu, I put this on the side while prioritizing other stuff but now I came back to it with #1027. I believe by introducing `cache-min-negative-ttl` both this PR and the initial issue #970 are addressed.

gthess:
#1027 is now merged, which replaces this PR with the new `cache-min-negative-ttl` option that can be used to affect the minimum TTL for negative answers. Thanks for bringing this up!
Closing.

PR #980, opened 2023-12-14
DoH: reject non-h2 early

### LINKED ISSUE #979: Improve errors for non-HTTP/2 DoH clients
Unbound requires HTTP/2 to serve DoH, which makes sense for performance reasons.

However, sometimes a client may be misconfigured to only send HTTP/1.1. For instance we recently ran into such a problem in Boulder: https://github.com/letsencrypt/boulder/pull/7215.

When that happens, Unbound's debug logs show:

    http2: session_recv from 192.168.4.146 failed, error: Received bad client magic byte string

And the client sees the connection closed abruptly with no response. In Go that looks like an `EOF` error. In curl it looks like:

    curl: (52) Empty reply from server

It would be nice to have more detailed information on what went wrong. For instance, on Unbound's side, rather than logging about the client magic string, we could notice that the client did not send `h2` as one of its ALPN protocols, and thus the client doesn't support HTTP/2.

Also, Unbound could support HTTP/1.1 handshakes to get as far as serving a HTTP 400 with an appropriate error message saying "Unbound requires HTTP/2 for DoH but client only supports HTTP/1.1."

jsha:
Previously, non-h2 connections would be accepted, and then error out with a verbose message "http2: session_recv from ____ failed, error: Received bad client magic byte string". Instead, we can detect absence of h2 support at connection time and reject with a clearer verbose message.

Partial fix for #979

### DIFF
--- a/util/netevent.c
+++ b/util/netevent.c
@@ -1748,6 +1748,9 @@ ssl_handshake(struct comm_point* c)
 			/* connection upgraded to HTTP2 */
 			c->tcp_do_toggle_rw = 0;
 			c->use_h2 = 1;
+		} else {
+			verbose(VERB_ALGO, "client doesn't support HTTP/2");
+			return 0;
 		}
 	}
 #endif


wcawijngaards (approved):
The commit has clear error reporting and this improves the log output. Code also looks fine.

wcawijngaards:
The fix makes a nicer error output, so that is a nice improvement! It is merged into the code repository.

PR #987, opened 2024-01-05
skip edns frag retry if advertised udp payload size is not smaller

borisVanhoof:
Hi,

If a serviced query is in UDP_EDNS_FRAG mode, and EDNS_ADVERTISED_SIZE is 1232 (the default) or less, then the retry will have the same edns udp payload size with the same result.

This patch will, if the advertised udp payload size is not smaller, increase the retry count and continue until answer or no more targets to forward to.

I used this conf to test and the following firewall rule to drop responses and cause timeout net event.
```
$ cat example.conf 
server:
    username: ""
    chroot: ""
    directory: "/tmp"
    do-daemonize: no
    use-syslog: no
    module-config: "iterator"
    ip-freebind: yes
    interface: lo
    infra-keep-probing: yes

forward-zone:
    name: "."
    forward-addr: 8.8.8.8

$ sudo iptables-legacy -I INPUT -p udp -s 8.8.8.8 --sport 53 -j DROP
```
I have added pre and post patch pcap files in [pcapfiles.tar.gz](https://github.com/NLnetLabs/unbound/files/13841993/pcapfiles.tar.gz). The pre file has 9 (I was expecting 10) requests to 8.8.8.8 and the post file has the expected 5 requests before servfail.

Kind regards and a happy new year 2024!

### DIFF
--- a/services/outside_network.c
+++ b/services/outside_network.c
@@ -2820,6 +2820,25 @@ serviced_perturb_qname(struct ub_randstate* rnd, uint8_t* qbuf, size_t len)
 	}
 }
 
+static uint16_t
+serviced_query_udp_size(struct serviced_query* sq, enum serviced_query_status status) {
+	uint16_t udp_size;
+	if(status == serviced_query_UDP_EDNS_FRAG) {
+		if(addr_is_ip6(&sq->addr, sq->addrlen)) {
+			if(EDNS_FRAG_SIZE_IP6 < EDNS_ADVERTISED_SIZE)
+				udp_size = EDNS_FRAG_SIZE_IP6;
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
The code is neat and well written. The reuse of the function makes is easier to read. Moving to retry at a smaller size that then is going to be the same value as before, is not useful for resolving the query. So this code change looks good to have.

wcawijngaards:
Thank you for the patch! That should avoid unnecessary retries, since the flag day changes made the default edns size drop from 4096 to 1232, the size is not smaller in that retry any more. The code has been merged into the code repository.

borisVanhoof:
Thank you.

PR #990, opened 2024-01-13
NAT64 support: put logging inside if

momoka0122y:
The logging code was not inside an if statement causing a log even when not needed.


gthess:
Is that from log file observation or reading the code?
`log_name_addr()` will eventually call `verbose()` which will test for the exact same thing before logging:
https://github.com/NLnetLabs/unbound/blob/3d95cef08c46b654d9605d196ec692ded63503bd/util/log.c#L356-L360
Are you maybe using `-v` when running Unbound? Any 'v' in the command line is increasing the already configured verbosity.

momoka0122y:
Sorry, this was only by reading the code.
I should have checked the log file.

gthess:
No problem! I am closing this as non-issue.

PR #993, opened 2024-01-16
Update b.root-servers.net also in example config file

pemensik:
Addition to commit a8739bad76d4d179290627e989c7ef236345bda6, which updated only address specified in code. But addresses provided in example configuration were not updated, I think they should be updated too.


wcawijngaards (approved):
Those are correct addresses for a better example config. The change looks good.

wcawijngaards:
Thank you for the fixup! The new address is committed to the example config file in the code repository.

PR #999, opened 2024-01-26
Search for protobuf-c with pkg-config

NickCao:
This is easier to use than `--with-protobuf-c=`.


wcawijngaards (approved):
This change looks like it is good code. Useful to detect more easily the protobuf-c location.

wcawijngaards:
Thanks for the additional check and that may make it easier to use the configure script! The script snippet is merged into the code repository and after a rerun of autoconf part of the configure script.

PR #1015, opened 2024-02-15
Fast reload

wcawijngaards:
This implements fast_reload. It reloads the unbound server and does not stop the processing of queries. So that service of queries remains uninterrupted by the reload process.

It can be used with `unbound-control fast_reload`. This then reloads the server. Cache is kept from before the reload to after. Also existing queries are kept, and the state machines continue to work on them after the reload. Only some options are changed, global options from the config structure, and forwards, and stubs.

The fast_reload works by creating a thread that reads config, and allocates memory. Then it quickly halts the other threads, updates the config and swaps the trees of forwards and stubs in place, and continues the other threads. A test with thousands of forwards, has a reload time of 0.027s with 0.0003s spent in halting the other threads.


gthess:
#1042 was based on this PR which is already merged, closing this.

PR #1027, opened 2024-03-12
Introduce 'cache-min-negative-ttl' option

gthess:
This will bound the minimum TTL for negative answers overriding `cache-min-ttl`, if set, for negative answers.
Setting `cache-min-negative-ttl: 1` while `cache-min-ttl` is also set would allow for upstream negative replies to not be capped while regular replies would be bounded to `cache-min-ttl`.
Related to #970 and #978.


wcawijngaards (approved):
The code change looks good! The extra option complements the existing options for cache TTLs.

PR #1030, opened 2024-03-15
Persist the openssl and expat directories for repeated windows builds

gthess:
This was lingering on my local repository.
Not really needed with current hardware but it can save time for repeated builds while debugging.


wcawijngaards (approved):
This looks fine. Could be useful for debugging.

PR #1041, opened 2024-04-03
Stub and Forward unshare

### LINKED ISSUE #1038: fatal error: Could not initialize thread / error: reading root hints
Hi!
**Describe the bug**
**Sporadic** fatal errors possible at unbound start if the `root-hints` option is used.
The server may become **unresponsive** as a result.

The copy of the https://www.internic.net/domain/named.root is used for root-hints.
Errors seen (can occur either during one start attempt or be limited to one error per unsuccessful attempt):
`error: reading root hints /root.hints 19:1: Empty label`
`error: reading root hints /root.hints 2:37: Syntax error, could not parse the RR's type`
`error: reading root hints /root.hints 2:9: Syntax error, could not parse the RR's type`
`error: reading root hints /root.hints 8:14: Syntax error, could not parse the RR's class`
`error: reading root hints /root.hints 13:15: Conversion error, ip6 addr expected`
and so on..
so it ends up with:
`error: Could not set root or stub hints` and `fatal error: Could not initialize thread` (for one or more threads)

The sporadic nature of the errors and their content (error in reading the record type in a comment line, mismatch between the record type and the address type) suggest that there may be a thread race when reading a file?

**To reproduce**
1. Use `root-hints` option
2. Сreate conditions for periodic unbound restart
3. Wait

**Expected behavior**

**System:**
 - Unbound version: **1.19.3**
 - OS: **FreeBSD 13.2-RELEASE-p10**
 - `unbound -V` output:
```
 --with-libexpat=/usr/local --with-ssl=/usr/local --enable-dnscrypt --disable-dnstap --with-libnghttp2 --with-dynlibmodule --enable-ecdsa --disable-event-api --enable-gost --with-libevent --with-pythonmodule=yes --with-pyunbound=yes ac_cv_path_SWIG=/usr/local/bin/swig LDFLAGS=-L/usr/local/lib --disable-subnet --disable-tfo-client --disable-tfo-server --with-pthreads --prefix=/usr/local --localstatedir=/var --mandir=/usr/local/share/man --infodir=/usr/local/share/info/ --build=amd64-portbld-freebsd13.2
Linked libs: libevent 2.1.12-stable (it uses kqueue), OpenSSL 3.0.13 30 Jan 2024
Linked modules: dns64 python dynlib respip validator iterator
DNSCrypt feature available

```

**Additional information**
not yet

**Thanks!**

cc @fichtner

wcawijngaards:
This merges the commits that unshare the stub and forward structures. This means they are locked and present once on the server. During a reload, then, then hints and also forwards are read only once, and then they are used with locks by the threads.

It fixes #1038 .

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -1278,7 +1278,7 @@ daemon.lo daemon.o: $(srcdir)/daemon/daemon.c config.h $(srcdir)/daemon/daemon.h
  $(srcdir)/util/edns.h $(srcdir)/services/listen_dnsport.h $(srcdir)/services/cache/rrset.h \
  $(srcdir)/services/cache/infra.h $(srcdir)/util/rtt.h $(srcdir)/services/localzone.h \
  $(srcdir)/services/authzone.h $(srcdir)/services/mesh.h $(srcdir)/services/rpz.h $(srcdir)/respip/respip.h \
- $(srcdir)/util/random.h $(srcdir)/util/tube.h $(srcdir)/util/net_help.h $(srcdir)/sldns/keyraw.h
+ $(srcdir)/util/random.h $(srcdir)/util/tube.h $(srcdir)/util/net_help.h $(srcdir)/sldns/keyraw.h $(srcdir)/iterator/iter_fwd.h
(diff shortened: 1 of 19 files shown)


gthess (approved):
This looks good and ready to merge!

PR #1043, opened 2024-04-09
Add loongarch support

xiaoxiaoafeifei:
1. Update config.guess(2024-01-01) and config.sub(2024-01-01) to solve the following issue:
      "configure: error: cannot guess build type; you must specify one"
  2. Add loongarch to little endian support


gthess (approved):
Verified the config.guess and config.sub files from the upstream.
Thanks for adding support for the loongarch!

PR #1048, opened 2024-04-15
Update ax_pkg_swig.m4 and ax_pthread.m4

pemensik:
Use vanilla m4 files with known source. Prepared for possible removal at build time if the system already has autoconf-archive source present. Switch to AX_PKG_SWIG macro for versioned or unversioned swig detection.

### DIFF
--- a/ax_pkg_swig.m4
+++ b/ax_pkg_swig.m4
@@ -1,39 +1,43 @@
 # ===========================================================================
-#              http://autoconf-archive.cryp.to/ac_pkg_swig.html
+#       https://www.gnu.org/software/autoconf-archive/ax_pkg_swig.html
 # ===========================================================================
 #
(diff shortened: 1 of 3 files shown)


pemensik:
Updates:

- ax_pthread: https://www.gnu.org/software/autoconf-archive/ax_pthread.html
- ax_pkg_swig: https://www.gnu.org/software/autoconf-archive/ax_pkg_swig.html

Source of files: autoconf-archive-2023.02.20-2.fc39.noarch

wcawijngaards (approved):
This updates the swig and pthread configure macros, the m4 scripts are the same as the source versions. That should be good, as Unbound then has newer scripts!

wcawijngaards:
Thanks for the update! Added the change and re-ran autoconf to create configure with the new sources.

PR #1053, opened 2024-04-22
Remove child delegations from cache when grandchild delegations are returned from parent

gthess:
An expired delegation could cause delegation invalidation for its descendants.
When a parent replies with a grandchild delegation, an expired child delegation may never be updated and cause all sub queries to end up in the parent.
With this PR when a gandchild delegation response is returned, child delegations up to the parent are removed from the cache to try and avoid this.

### DIFF
--- a/iterator/iterator.c
+++ b/iterator/iterator.c
@@ -52,6 +52,7 @@
 #include "iterator/iter_priv.h"
 #include "validator/val_neg.h"
 #include "services/cache/dns.h"
+#include "services/cache/rrset.h"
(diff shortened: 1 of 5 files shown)


wcawijngaards (approved):
The code looks good. The code removes the expired NS entries for multilabel delegations, from the intermediate labels, so that the ghost domain fixup does not trip on them as expired, but never removed, entries. This keeps the expired NS records in a way to detect that parental delegations need to be rechecked.

gthess:
Thanks!

PR #1054, opened 2024-04-24
Update locking management for iter_fwd and iter_hints methods.

gthess:
Move most of the locking management to iter_fwd and iter_hints methods.
The caller still has the ability to handle its own locking, if desired, for atomic operations on sets of different structs.

### DIFF
--- a/daemon/cachedump.c
+++ b/daemon/cachedump.c
@@ -839,6 +839,7 @@ int print_deleg_lookup(RES* ssl, struct worker* worker, uint8_t* nm,
 	char b[260];
 	struct query_info qinfo;
 	struct iter_hints_stub* stub;
+	int nolock = 0;
 	regional_free_all(region);
 	qinfo.qname = nm;
 	qinfo.qname_len = nmlen;
@@ -851,8 +852,7 @@ int print_deleg_lookup(RES* ssl, struct worker* worker, uint8_t* nm,
 		"of %s\n", b)) 
 		return 0;
 
-	lock_rw_rdlock(&worker->env.fwds->lock);
-	dp = forwards_lookup(worker->env.fwds, nm, qinfo.qclass);
+	dp = forwards_lookup(worker->env.fwds, nm, qinfo.qclass, nolock);
 	if(dp) {
 		if(!ssl_printf(ssl, "forwarding request:\n")) {
 			lock_rw_unlock(&worker->env.fwds->lock);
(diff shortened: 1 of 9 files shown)


daemon/cachedump.c
```diff
@@ -851,8 +852,7 @@ int print_deleg_lookup(RES* ssl, struct worker* worker, uint8_t* nm,
 		"of %s\n", b)) 
 		return 0;
 
-	lock_rw_rdlock(&worker->env.fwds->lock);
-	dp = forwards_lookup(worker->env.fwds, nm, qinfo.qclass);
+	dp = forwards_lookup(worker->env.fwds, nm, qinfo.qclass, nolock);
```
wcawijngaards, line 855: Both forwards_lookup and hints_lookup_stub return a pointer to the delegpt that is part of the forward and hints tree. And that has to be used while the lock is held, and so the nolock option, where it routine releases the lock and then returns a pointer to contents in the tree, is not good for thread safety. The routine has to stay as it is and protect the accesses, that print the delegation point, by holding the lock.

wcawijngaards: But it does not release the lock when the return value is not NULL. So that solves it.

pythonmod/interface.i
```diff
@@ -1415,8 +1415,8 @@ struct delegpt* dns_cache_find_delegation(struct module_env* env,
         int noexpiredabove, uint8_t* expiretop, size_t expiretoplen);
 int iter_dp_is_useless(struct query_info* qinfo, uint16_t qflags,
         struct delegpt* dp, int supports_ipv4, int supports_ipv6, int use_nat64);
-struct iter_hints_stub* hints_lookup_stub(struct iter_hints* hints,
-        uint8_t* qname, uint16_t qclass, struct delegpt* dp);
+struct iter_hints_stub* hints_lookup_stub_foo(struct iter_hints* hints,
```
wcawijngaards: Fixup function name edit.
```suggestion
struct iter_hints_stub* hints_lookup_stub(struct iter_hints* hints,
```

pythonmod/interface.i
```diff
@@ -1455,14 +1456,13 @@ struct delegpt* find_delegation(struct module_qstate* qstate, char *nm, size_t n
             dname_str((uint8_t*)nm, b);
             continue;
         }
-        lock_rw_rdlock(&qstate->env->hints->lock);
-        stub = hints_lookup_stub(qstate->env->hints, qinfo.qname, qinfo.qclass, dp);
+        stub = hints_lookup_stub_foo(qstate->env->hints, qinfo.qname,
```
wcawijngaards: Fixup function name edit.
```suggestion
        stub = hints_lookup_stub(qstate->env->hints, qinfo.qname,
```

wcawijngaards (approved):
The code looks fine. It simplifies the locks a bit.

PR #1062, opened 2024-05-07
Fix potential overflow bug while parsing port in function cfg_mark_ports

xiaoxiaoafeifei:
Fix potential overflow bugs caused by the following situations:  use strtol function instead of atoi function
1. The string to be parsed may contain non-numeric characters.
2. The port parsed from the string may be negative.

### DIFF
--- a/util/config_file.c
+++ b/util/config_file.c
@@ -42,6 +42,7 @@
 #include "config.h"
 #include <ctype.h>
 #include <stdarg.h>
+#include <errno.h>
 #ifdef HAVE_TIME_H
 #include <time.h>
 #endif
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
Nice to check for more error conditions. I cannot spot where it would have had negative numbers from the string.

wcawijngaards:
Thank you for the contribution! The fixup commit moves a declaration before statements to avoid a warning, then avoids printing a null value on error, just in case, and adds a redundant check on array size, also just for extra certainty. The pull request should catch unknown characters in that string and also range errors and that is nice to have.

PR #1069, opened 2024-05-16
Fix unbound-control stdin commands for multi-process Unbounds

gthess:
This PR distributes commands that read stdin properly to other processes. These are `local_zones_remove`, `local_zones`, `local_datas_remove`, `local_datas`, `view_local_datas_remove` and `view_local_datas`.
`testdata/09-unbound-control.tdir` is updated to check the correctness on different scenarios. It now does it one line at a time but it can be updated to put several lines in the same distribute batch.

`dump_cache` and `load_cache` are no longer supported in multi-process operation because of confusing results in that scenario.

`testdata/remote-threaded.tdir` is removed as it is obsoleted by the updated `testdata/09-unbound-control.tdir`.


daemon/remote.c
```diff
@@ -633,6 +637,25 @@ static void send_ok(RES* ssl)
...
+	int i;
+	if(!cmd || !ssl)
+		return;
+	/* skip i=0 which is me */
+	for(i=1; i<rc->worker->daemon->num; i++) {
+		worker_send_cmd(rc->worker->daemon->workers[i],
+			worker_cmd_remote);
+		if(!tube_write_msg(rc->worker->daemon->workers[i]->cmd,
+			(uint8_t*)cmd, strlen(cmd)+1, 0)) {
+			ssl_printf(ssl, "error could not distribute cmd\n");
```
wcawijngaards:
```suggestion
			(void)ssl_printf(ssl, "error could not distribute cmd\n");
```

wcawijngaards (approved):
This looks like it would fix consistency for config changes, by distributing the content to the worker processes. The code looks fine.

PR #1070, opened 2024-05-16
Fix rtt assignement for low values of infra-cache-max-rtt

gthess:
When a server is deemed still usable keep the same logic (a useful value outside of the RTT_BAND) also for small configured values of infra-cache-max-rtt.


wcawijngaards (approved):
This looks good, so that timeouted servers stay candidate for selection when low max rtt is chosen.

gthess:
Thanks!

PR #1078, opened 2024-05-29
Only check old pid if no username

vopatek:
Do as the comment says and only check old pid if there is no username configured.


wcawijngaards (approved):
Code is good, and fixes an oversight.

wcawijngaards:
Thank you for the fix! The comment was correct and indicates what should happen, the bugfix helps detect past failures.

PR #1080, opened 2024-05-31
AddressSanitizer detection in tdir tests and memory leak fixes

gthess:
Checks the output of the tdir tests for any Address Sanitizer errors and fails the test.
This will increase visibility of memory leaks on different test cases if compiled with `--fsanitize=address`.

It also fixes a memory leak when `reload_keep_cache` is used **AND** `num-threads` changes for the reload.

It also fixes a memory leak when exiting the test program `unbound-dnstap-socket` that now creates false negatives during testing. For this `unbound-dnstap-socket` introduces a unit test that I run under `02-unittest.tdir`. Not 100% if this place makes sense or it needs to be relocated.

### DIFF
--- a/testdata/02-unittest.tdir/02-unittest.test
+++ b/testdata/02-unittest.tdir/02-unittest.test
@@ -7,7 +7,7 @@
 . ../common.sh
 PRE="../.."
 get_make
-(cd $PRE ; $MAKE unittest; $MAKE lock-verify)
+(cd $PRE ; $MAKE unittest; $MAKE lock-verify; $MAKE unbound-dnstap-socket)
 
 if test -f $PRE/unbound_do_valgrind_in_test; then
 	do_valgrind=yes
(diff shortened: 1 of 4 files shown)


testdata/02-unittest.tdir/02-unittest.test
```diff
@@ -16,11 +16,16 @@ else
 fi
 VALGRIND_FLAGS="--leak-check=full --show-leak-kinds=all"
 
+## START -- Loop over unit tests
+##
+for unit_cmd in "unittest" "unbound-dnstap-socket -c"; do
```
wcawijngaards: This builds unbound-dnstap-socket also when dnstap is not enabled, that seems to be harmless. But it also runs the '-c' test on it, when there is no enable-debug, this is going to fail non-debug compile tests.

wcawijngaards (approved):
Looks okay, nice to catch the memory leaks. It needs to skip parts when debug is not enabled, or possibly also skip parts when dnstap is not enabled.

testdata/02-unittest.tdir/02-unittest.test
gthess: Great catch! I believe I solved it with the latest commit.

wcawijngaards: Yes, that looks fixed.

PR #1087, opened 2024-06-07
Overload `local_data_remove` to support removing specific records

rcmcdonald91:
Here we overload the `local_data_remove` control command to support deleting specific records. Curently, this command deletes all records for a given zone. The modification works by attempting to parse the command argument first as a complete record and then as just a domain name, if the first attempt failed.

This preserves the command's behavior, while also supporting removing specific records from the zone tree.

### DIFF
--- a/daemon/remote.c
+++ b/daemon/remote.c
@@ -1,6 +1,7 @@
 /*
  * daemon/remote.c - remote control for the unbound daemon.
  *
+ * Copyright (c) 2024, Rubicon Communications, LLC ("Netgate"). All rights reserved.
  * Copyright (c) 2008, NLnet Labs. All rights reserved.
  *
  * This software is open source.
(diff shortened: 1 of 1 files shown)


rcmcdonald91:
@wcawijngaards can I get some eyes on this? Thanks!

rcmcdonald91:
@gthess

tkreagan:
@wcawijngaards hi - can we get some feedback or an update on this?  I would also welcome support for this feature, and am interested in understanding the project's position on it.

Another possible approach I considered was creating a local_data_remover command (same naming idea as local_datas) that was local_data_remove but scoped to a single record.  Would that be more palatable rather than overloading an existing command?

wcawijngaards:
Hi, I appreciate the code and also the additional feature. The code has a copyright statement, and the ownership of the code, we want to have that with us. And we want to have the code as open source under the BSD license. Since this is in the main codebase (as opposed to 'contrib' or 'compat' or a library), we would like to have the code contributed to use under the BSD license. (That is the license we use for this open source project). If you agree with that, then I think removing the added `Copyright` line from the suggested change would be way to make this look like any other contribution that we receive. So one of the reasons we want that is to avoid some shared ownership of code and having to communicate with these partners, or find them, for making decisions, down the line.

rcmcdonald91:
> If you agree with that, then I think removing the added `Copyright` line from the suggested change would be way to make this look like any other contribution that we receive. So one of the reasons we want that is to avoid some shared ownership of code and having to communicate with these partners, or find them, for making decisions, down the line.

This is not an issue.

PR #1090, opened 2024-06-14
Cookie secret file

### LINKED ISSUE #1088: RFC 9018 compliance (cookie secret rollover)
**Describe the bug**

RFC 9018 further clarifies the DNS cookie feature introduced with RFC 7873. Unbound is capable of responding to downstream requests with a cookie via the `answer-cookie` and the `cookie-secret` option introduced with https://github.com/NLnetLabs/unbound/pull/762

There are further options like for the `wait-limit` (introduced with 1.20) that make use of Unbound's support of DNS cookies.

However, Unbound does not provide any way to perform a three stage rollover for updating the cookie secret as described in [RFC 9018 Section 5](https://www.rfc-editor.org/rfc/rfc9018#name-updating-the-server-secret).

**Additional information**

NSD provides the option to rotate the cookie secret via `nsd-control` arguments.

See:
- `add_cookie_secret`
- `drop_cookie_secret`
- `activate_cookie_secret`

Having some configuration option like `cookie-secret-staging` that one can use to swap the active/staging secret for the 3 stage rollover might work as well.

On a minor notice, neither RFC 9018 nor 7873 are listed on [Unbound's RFC compliance list](https://unbound.docs.nlnetlabs.nl/en/latest/reference/rfc-compliance.html).

Edit: I just noticed that cookie secret rollover was a `todo` in the PR that introduced its feature. However, after merging the PR about 1 year ago it looks like it was never picked up again. With more features (like `wait-limit`) introducing options that rely on downstream cookies, it might be worthwhile taking a look at the secret rollover aspect of DNS cookie usage.

wcawijngaards:
This creates a new option to store the cookies in a file and manage them with remote control.

The `cookie-secret-file: "unbound_cookiesecrets.txt"` option can be used with the unbound-control `add_cookie_secret`, `drop_cookie_secret`, `activate_cookie_secret` and `print_cookie_secrets` commands. Code has been used from the NSD implementation of it.

This fixes #1088 .


daemon/remote.c
```diff
@@ -3193,6 +3194,211 @@ do_rpz_disable(RES* ssl, struct worker* worker, char* arg)
...
+		UNBOUND_COOKIE_SECRET_SIZE ) {
+		lock_basic_unlock(&cookie_secrets->lock);
+		explicit_bzero(secret, UNBOUND_COOKIE_SECRET_SIZE);
+		explicit_bzero(arg, strlen(arg));
+		(void)ssl_printf(ssl, "invalid cookie secret: parse error\n");
+		(void)ssl_printf(ssl, "please provide a 128bit hex encoded secret\n");
+		return;
+	}
+	/* Only the worker 0 writes to file, the others update state. */
+	if(worker->thread_num == 0 && !cookie_secret_file_dump(ssl, worker)) {
```
gthess, line 3353: Do we need to write two times to file? Once for the current state and one after the addition? Why not just after the addition?
Also similar question for `do_drop_cookie_secret()`, `do_activate_cookie_secret()`.

gthess (changes requested):
Looks good!
Just a question about writing to file.
Some changes for the test and some documentation edits will follow in a PR; that's why the "request changes" status.

gthess:
If the attached PR (https://github.com/NLnetLabs/unbound/pull/1116) is merged here, then a squash and merge of this back to master should do it.

daemon/remote.c
wcawijngaards: NSD does this as well, https://github.com/NLnetLabs/nsd/blob/b88421be3b5654eed5e2c432d20172b72ea12052/remote.c#L2410 . This checks that the file is actually writable before making the change. It seems a useful safeguard, so I do not think it is an issue.

wcawijngaards:
Thank you for the review! Added the changes and squashed and merged it into the repo.

PR #1098, opened 2024-07-02
Ipset pf support

wcawijngaards:
This cleans up the code, after discussion with my colleague, we decided to rename the module calls differently. This makes that change. It also performs a merge with the recent code base.

Fixes #144 .


ipset/ipset.c
```diff
@@ -178,23 +272,24 @@ ipset_check_zones_for_rrset(struct module_env *env, struct ipset_env *ie,
...
 
-	mnl = (struct mnl_socket *)ie->mnl;
-	if (!mnl) {
-		/* retry to create mnl socket */
-		mnl = open_mnl_socket();
-		if (!mnl) {
+#ifdef HAVE_NET_PFVAR_H
+#else
+	if (!ie->dev) {
+		// retry to create mnl socket
```
gthess:
```suggestion
		/* retry to create mnl socket */
```

ipset/ipset.c
```diff
@@ -41,24 +51,95 @@ static int error_response(struct module_qstate* qstate, int id, int rcode) {
...
 
-	if (mnl_socket_bind(mnl, 0, MNL_SOCKET_AUTOPID) < 0) {
-		mnl_socket_close(mnl);
+	if (mnl_socket_bind(dev, 0, MNL_SOCKET_AUTOPID) < 0) {
+		mnl_socket_close(dev);
 		log_err("ipset: could not bind netfilter.");
 		return NULL;
 	}
-	return mnl;
+	return dev;
```
gthess:
```suggestion
	return (void *)dev;
```

gthess (commented):
Some nits in the suggestions.
But some additional points were raised during review:
- In the module function block I would prefer the _startup functions to be before the _init ones to match the perceived execution order (nit).
- Now that we have `modstack_(de)startup` I would like `modstack_(de)setup` to be renamed to `modstack_(de)init`, and `modstack_init` to be renamed to `modstack_something`? But then I think that the better names would be `modstack_call_(de)startup` and `modstack_call_(de)init`. Then `modstack_init` can just stay the same. That way we have distinct naming when modstack methods are for the modstack struct itself and when they are calling module functions.
- I think for the reordering of the modules during reload, we can block only if any module has a startup function, otherwise we don't care, right?

wcawijngaards:
Added a commit that fixes those review issues. The naming like `modstack_call_deinit` is nice. I have also added `modstack_free`. That is also used when it detects a change in module list that does not involve modules that need `startup`. The modstack deinit was called too often, during the first start as well, the `daemon->mods_inited` variable tracks that, so that deinit is called during reloads, but not when the server starts for the first time.

gthess (approved):
This looks good to me now, thanks!

PR #1111, opened 2024-07-26
Shumon's invalid query type EDE return

wtoorop:
As described in https://datatracker.ietf.org/doc/draft-ietf-dnsop-compact-denial-of-existence


shuque:
I approve! :)

PR #1116, opened 2024-08-01
Update cookie_file.tdir

gthess:
This removes the unshare and faketime dependencies for cookie_file testing.
Unbound already has unittests for the RFC scenarios and faketime interferes with AddressSanitizer.
The test case is also more robust with configured BADCOOKIE replies from Unbound.


wcawijngaards (approved):
Nice to have the test work with debuggers. The documentation changes are also nice.

PR #1132, opened 2024-08-30
b.root renumbering

loganaden:
https://b.root-servers.org/news/2023/05/16/new-addresses.html

Worked together with Jaykishan Muktawoa <jay@cyberstorm.mu>


wcawijngaards (approved):
These are the updated B root addresses. But this is a unit test file, it can contain a copy of an older version of the root zone for checking. It can be updated, of course.

PR #1140, opened 2024-09-15
Fix spelling mistake in comments

TochusC:
I noticed a spelling mistake in the comments. The term “chain of trust” was incorrectly written as “chainoftrust”. This change corrects the spelling to “chain of trust” which is the correct term used in English.


wcawijngaards:
Thank you for the spelling correction! It was spelled like that because the term is coined for use in DNSSEC processing.

PR #1143, opened 2024-09-23
Fix cache update when serve expired is used

### LINKED ISSUE #994: serve expired fails when DNSSEC is enabled
**Describe the bug**
Unbound's serv-expired feature seems to break when DNSSEC is enabled.

**To reproduce**
1. Use default Unbound configuration in ``` /etc/unbound/unbound.conf``` and ensure DNSSEC and serving stale is enabled. Also reduce cache max ttl for reproduce. See sample config below.  
```
        # Serve Stale enabled  (RFC 8767)
	serve-expired: yes
	serve-expired-ttl: 86400
	serve-expired-reply-ttl: 30
	serve-expired-client-timeout: 1000
       
       # DNSSEC enabled
        module-config: "validator iterator"
	trusted-keys-file: /etc/unbound/keys.d/*.key
	auto-trust-anchor-file: "/var/lib/unbound/root.key"
	prefetch-key: yes
	val-clean-additional: yes
	val-permissive-mode: no
       
       # lower cache max TTL to trigger the issue
       cache-max-ttl: 10
       
        # Misc configs 
        log-queries: yes
        log-servfail: yes
        prefetch: yes
	rrset-roundrobin: yes
	minimal-responses: yes
	qname-minimisation: no
        harden-glue: yes

```
2.  restart unbound ```unbound-control -c /etc/unbound/unbound.conf```
3.  Send a query this ensure the cache is populated for this query
```dig s3.amazonaws.com```
4.  Block upstream name servers. E.g. this iptable rule blocks all .COM servers which are in the 192/8 range
```
# Note: change -A to -D to remove blocks
/sbin/iptables -o eth0 -A OUTPUT -p udp -d 192.0.0.0/8 -j DROP
/sbin/iptables -o eth0 -A OUTPUT -p tcp -d 192.0.0.0/8 -j DROP
```
5. keep sending dig and observe that after several seconds it returns SERVFAIL
```
while true 
do 
        dig s3.amazonaws.com; 
        sleep 1;
done

i.e.
; <<>> DiG 9.11.4-P2-RedHat-9.11.4-26.P2.amzn2.13.5 <<>> s3.amazonaws.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 23877
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;s3.amazonaws.com.		IN	A

;; Query time: 1 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Tue Jan 16 16:03:18 UTC 2024
;; MSG SIZE  rcvd: 45

```

**Expected behavior**

Repeat steps from 1 - 5 with DNSSEC disabled. Change module-config to: 
``` module-config: "iterator"```

With DNSSEC disabled Unbound responds from stale cache.  This is confirmed by the 30 second TTL on answers. 

```
dig s3.amazonaws.com 

; <<>> DiG 9.11.4-P2-RedHat-9.11.4-26.P2.amzn2.13.5 <<>> s3.amazonaws.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 44358
;; flags: qr rd ra; QUERY: 1, ANSWER: 8, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
;; QUESTION SECTION:
;s3.amazonaws.com.		IN	A

;; ANSWER SECTION:
s3.amazonaws.com.	30	IN	A	16.182.68.56
s3.amazonaws.com.	30	IN	A	52.217.4.62
s3.amazonaws.com.	30	IN	A	52.216.94.205
s3.amazonaws.com.	30	IN	A	16.182.37.24
s3.amazonaws.com.	30	IN	A	52.216.209.112
s3.amazonaws.com.	30	IN	A	52.217.169.8
s3.amazonaws.com.	30	IN	A	52.217.202.128
s3.amazonaws.com.	30	IN	A	52.217.115.24

;; Query time: 1 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Tue Jan 16 16:03:26 UTC 2024
;; MSG SIZE  rcvd: 173
```

**System:**
 - Unbound version:  1.12.0 and  1.16.0
 - OS: ```Linux 5.10.201-191.748.amzn2.aarch64 GNU/Linux```
 - `unbound -V` output:

Both Unbound 1.12.0 and 1.160.0 seems to have the same issue.

```
/usr/sbin/unbound  -V
Version 1.12.0

Configure line: --prefix=/usr --sysconfdir=/etc --libdir=/usr/lib64 --with-libevent
Linked libs: libevent 2.0.21-stable (it uses epoll), OpenSSL 1.0.2k-fips  26 Jan 2017
Linked modules: dns64 respip validator iterator

BSD licensed, see LICENSE in source package for details.
Report bugs to unbound-bugs@nlnetlabs.nl or https://github.com/NLnetLabs/unbound/issues
```

```
/usr/local/sbin/unbound -V
Version 1.16.0

Configure line: --with-libevent
Linked libs: libevent 2.0.21-stable (it uses epoll), OpenSSL 1.0.2k-fips  26 Jan 2017
Linked modules: dns64 respip validator iterator

BSD licensed, see LICENSE in source package for details.
Report bugs to unbound-bugs@nlnetlabs.nl or https://github.com/NLnetLabs/unbound/issues
```

**Additional information**
* Turned off prefetching assuming it might be the same bug as https://github.com/NLnetLabs/unbound/issues/394 but the issue remains with ```prefetch: no``` and ```prefetch-key: no``` configs. 
* Verbose logging shows DNSSEC validation errors when attempting to contact .COM servers and SERVFAILs as a result. This could be related.

### PR DESCRIPTION
**Pull request by gthess**
By not replacing/evicting still usable expired records. Modules are forbidden to update the cache if their answer is DNSSEC unchecked or bogus and a valid (expired) entry already exists. Bogus replies from the validator are also discarded in favor of existing (expired) valid replies.

Fixes #994.

### DIFF
--- a/testdata/serve_expired_client_timeout_val_bogus.rpl
+++ b/testdata/serve_expired_client_timeout_val_bogus.rpl
@@ -1,0 +1,317 @@
+; config options
+; The island of trust is at example.com
+server:
+	trust-anchor: "example.com.    3600    IN      DS      2854 3 1 46e4ffc6e9a4793b488954bd3f0cc6af0dfb201b"
+	val-override-date: "20070916134226"
+	target-fetch-policy: "0 0 0 0 0"
+	qname-minimisation: "no"
+	fake-sha1: yes
+	trust-anchor-signaling: no
+	minimal-responses: no
+
+	serve-expired: yes
+	serve-expired-client-timeout: 1
+	serve-expired-reply-ttl: 123
+	ede: yes
+	ede-serve-expired: yes
+
+	# No need for AAAA nameserver queries
+	do-ip6: no
+
+stub-zone:
+	name: "."
+	stub-addr: 193.0.14.129 	# K.ROOT-SERVERS.NET.
+CONFIG_END
+
+SCENARIO_BEGIN Test serve-expired with client-timeout and bogus answer
+; Scenario overview:
+; - query for www.example.com. IN A
+; - check the answer
+; - wait for the record to expire
+; - (upstream now has a bogus response)
+; - query again for www.example.com. IN A
+; - check that we get the expired valid response instead
+; - query once more
+; - (upstream has the valid response again)
+; - check that we get the immediate expired valid response
+; - (the prefetch query updates the cache with the valid response)
+; - query one last time
+; - check that we get the immediate valid cache response; upstream does not have an answer at this moment
+
+;;
+;; K.ROOT-SERVERS.NET.
+;;
+RANGE_BEGIN 0 100
+	ADDRESS 193.0.14.129 
+    ENTRY_BEGIN
+    MATCH opcode qtype qname
+    ADJUST copy_id
+    REPLY QR NOERROR
+    SECTION QUESTION
+    . IN NS
+    SECTION ANSWER
+    . IN NS	K.ROOT-SERVERS.NET.
+    SECTION ADDITIONAL
+    K.ROOT-SERVERS.NET.	IN	A	193.0.14.129
+    ENTRY_END
+
+    ENTRY_BEGIN
+    MATCH opcode
+    ADJUST copy_id copy_query
+    REPLY QR NOERROR
+    SECTION QUESTION
+    www.example.com. IN A
+    SECTION AUTHORITY
+    com.	IN NS	a.gtld-servers.net.
+    SECTION ADDITIONAL
+    a.gtld-servers.net.	IN 	A	192.5.6.30
+    ENTRY_END
+RANGE_END
+
+;;
+;; a.gtld-servers.net.
+;;
+RANGE_BEGIN 0 100
+	ADDRESS 192.5.6.30
+    ENTRY_BEGIN
+    MATCH opcode qtype qname
+    ADJUST copy_id
+    REPLY QR NOERROR
+    SECTION QUESTION
+    com. IN NS
+    SECTION ANSWER
+    com.    IN NS   a.gtld-servers.net.
+    SECTION ADDITIONAL
+    a.gtld-servers.net.     IN      A       192.5.6.30
+    ENTRY_END
+
+    ENTRY_BEGIN
+    MATCH opcode
+    ADJUST copy_id copy_query
+    REPLY QR NOERROR
+    SECTION QUESTION
+    www.example.com. IN A
+    SECTION AUTHORITY
+    example.com.	IN NS	ns.example.com.
+    SECTION ADDITIONAL
+    ns.example.com.		IN 	A	1.2.3.4
+    ENTRY_END
+RANGE_END
+
+;;
+;; ns.example.com. with generic valid data
+;;
+RANGE_BEGIN 0 100
+	ADDRESS 1.2.3.4
+    ENTRY_BEGIN
+    MATCH opcode qtype qname
+    ADJUST copy_id
+    REPLY QR NOERROR
+    SECTION QUESTION
+    example.com. IN NS
+    SECTION ANSWER
+    example.com.    IN NS   ns.example.com.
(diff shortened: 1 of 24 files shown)

### CODE REVIEW

File: validator/validator.c [hunk 1]
```diff
@@ -2449,10 +2451,51 @@ processFinished(struct module_qstate* qstate, struct val_qstate* vq,
...
+			if(rep && rep->security > sec_status_bogus &&
+				(!qstate->env->cfg->serve_expired_ttl ||
+				 qstate->env->cfg->serve_expired_ttl_reset ||
+				*qstate->env->now <= rep->serve_expired_ttl)) {
+				verbose(VERB_ALGO, "validation failed but "
+					"previously cached valid response "
+					"exists; set serve-expired-norec-ttl "
+					"for response in cache");
+				rep->serve_expired_norec_ttl = NORR_TTL +
+					*qstate->env->now;
```
wcawijngaards, line 2470: Should this perhaps be the bogus ttl, `val-bogus-ttl`, since there is a validation failure perhaps it could wait for that before it retries validation.
```suggestion
				rep->serve_expired_norec_ttl = ve->bogus_ttl +
					*qstate->env->now;
```

testdata/serve_expired_client_timeout_val_bogus.rpl
```diff
@@ -0,0 +1,317 @@
...
+ENTRY_END
+
+STEP 40 QUERY
+ENTRY_BEGIN
+REPLY RD DO
+SECTION QUESTION
+www.example.com. IN A
+ENTRY_END
+
+; immediate cached answer because upstream is valid again
```
wcawijngaards, line 273: Is the validation of the content, after it was bogus, not waiting for the serve_expired_norec_ttl? Then it would reply with cached contents for that time, but not attempt resolution?

testdata/serve_expired_val_bogus.rpl
```diff
@@ -0,0 +1,316 @@
...
+ENTRY_END
+
+STEP 40 QUERY
+ENTRY_BEGIN
+REPLY RD DO
+SECTION QUESTION
+www.example.com. IN A
+ENTRY_END
+
+; this is still the immediate cache response because the previous upstream response was bogus
```
wcawijngaards, line 273: Similarly, is this entry not waiting for the serve_expired_norec_ttl, and thus does not attempt to perform validation, after returning the expired cache response?

wcawijngaards (approved):
Code looks good. It is nice to keep the better validated answers in cache when there are lookup failures.

validator/validator.c
gthess: Maybe, and then clients will get the immediate cache response for the following "bogus_ttl" seconds, instead of waiting for resolution/validation first.
Validation will still happen in each such case since serve_expired_norec_ttl is essentially turning serve-expired-client-timeout configurations to just serve-expired (immediate cache answers and then resolution).

wcawijngaards: I see, that resolved also the two other questions I had. This behaviour should maybe be noted at the header file where the serve_expired_norec_ttl variable is declared. In that case the bogus-ttl is not really that applicable, and the time change suggested does not make that much sense to apply.

gthess: Just added more text on the header file

wcawijngaards: That is nice to read up, thank you for the explanatory text.

PR #1157, opened 2024-10-17
Fix heap corruption when calling ub_ctx_delete in Windows

lnzhu:
ub_ctx_delete in libunbound causes heap corruption exception in Windows OS. This issue is only for Windows and discovered when using the lastest libunbound Windows library published [here](https://nlnetlabs.nl/downloads/unbound/unbound-1.21.1.zip).
* ub_ctx_delete stops worker thread by [ub_stop_bg](https://github.com/NLnetLabs/unbound/blob/9a63db344eb859d32d04dcabcf0cd74fe00be76f/libunbound/libunbound.c#L347).
* When the workder thread exits, [tube_remove_bg_listen](https://github.com/NLnetLabs/unbound/blob/9a63db344eb859d32d04dcabcf0cd74fe00be76f/libunbound/libworker.c#L363) is called for `ctx->qq_pipe` and the memory of tube->ev_listen is released in ub_winsock_unregister_wsaevent but **tube->ev_listen is never set to NULL**.
* After worker thread stops, [tube_delete](https://github.com/NLnetLabs/unbound/blob/9a63db344eb859d32d04dcabcf0cd74fe00be76f/libunbound/libunbound.c#L382) is called for `tx->qq_pipe` in ub_ctx_delete.
* tube_delete calls [tube_remove_bg_listen](https://github.com/NLnetLabs/unbound/blob/9a63db344eb859d32d04dcabcf0cd74fe00be76f/util/tube.c#L559C2-L559C23) again and it tried to free tube->ev_listen again which leads to heap corruption.
* I tested multiple older versions (back to 2019) of libunbound Windows library and they all have the same issue.
* A cleaner way to fix this issue might be setting NULL inside ub_winsock_unregister_wsaevent but it will need to change ub_winsock_unregister_wsaevent API and pass `ub_event**` instead, which will bring additional changes in different places.


gthess:
Thanks, this looks good! But I don't think the `#if defined` is required. I see why you have it there, but always setting `tube->ev_listen = NULL` seems fine regardless if `USE_MINI_EVENT` is defined or not. Unless I miss something.

lnzhu:
Thanks for reviewing! Yes, `#if defined(USE_MINI_EVENT)` is not strictly needed in current code. But `tube->ev_listen` is deleted in  `ub_winsock_unregister_wsaevent`  only if `USE_MINI_EVENT` is defined. I think it might be safer and future-proof to have the check before calling `ub_winsock_unregister_wsaevent`.

gthess:
It is indeed only freed when `USE_MINI_EVENT` is defined.
For the other cases (when not defined, or when `ub_event_pluggable.c` is used) setting `tube->ev_listen = NULL` always after the `ub_winsock_unregister_wsaevent()` call conveys the correct intent that this listener event is no more.

I also prefer that all the `USE_MINI_EVENT` `#ifdef`s are handled in `ub_event(_pluggable).c` and not exposed in `tube.c`.

lnzhu:
removed `#if defined` in the last commit

gthess (approved):
Thanks for this!

PR #1159, opened 2024-10-23
Stats for discard-timeout and wait-limit

wcawijngaards:
The new discard-timeout and wait-limit options have been accounted as requestlist_exceeded in the statistics. But this mixes them with the request exceed number. The change makes their own statistics counters for them. The new statistics counters are `num.queries_discard_timeout` and `num.queries_wait_limit`.


gthess (approved):
Looks good, useful statistics to differentiate between performance drops and configuration drops of packets.

PR #1162, opened 2024-10-25
[Feat] Unbound: Support per-rule (local-zone) ipset targets

EngineersBox:
## Problem

In the existing workflow and configuration, Unbound requires an ipset to be specified globally and referenced where necessary in each `local-zone` declaration.
This is very limiting as it does not allow delineation of domains operating in different contexts, i.e. utilising ipsets for each port or protocol.
Having all rules bundled into a single ipset implies uniformity in access (egress/ingress) when bound to an ipset, potentially causing unwanted firewall access for domains over ports/interfaces/etc that are not desired.

For example, using the follwing config to populate an ipset for two domains where `https_example.com` operates over https port 443 and `amqp_domain.net` operates over 5431:

```conf
server:
    local-zone: "https_example.com." ipset
    local-zone: "amqp_domain.amqp." ipset

ipset:
    name-v4: "test_set"
```

Causes issues when trying to create matching iptables rules to allow to distinct ports on each domain:

```bash
ipset -A OUTPUT -m tcp -p tcp --dport 443 -m set --match-set test_set dst -J ALLOW
ipset -A OUTPUT --dport 5431 -m set --match-set test_set dst -J ALLOW
```

As both `https_example.com` and `amqp_domain.amqp` will be allowed on port 443 and 5431.

## Changes

This PR does the following:
* Supports per-rule ipset targets
* Allows for ipset entries to have expiry set based on RRSet entry TTL
* Ensures backwards compatibility with global ipset declaration
* Handles netlink ACK responses appropriately to check for errors (including permissions)

The new `local-zone` declaration format is: `lcoal-zone: <domain> ipset [<protocol> <set name> <ttl | no-ttl>]`.
The last parameters are still optional, should the user still be using the global ipset declaration.
Global declarations will be translated into per-rule definitions internally for simplicity, but have no difference in their behaviour.

The `protocol` field is designed for network layer protocols (i.e. ipv4/ipv6/arpa/etc), it only supports `ipv4` and `ipv6` values currently inline with the existing supported protocols, appropriately logging an error when encountering an invalid value.

### Example Configuration

An example config in this new setup is:

```
server:
    local-zone: "https_example.com." ipset ipv4 "https_set" ttl
    local-zone: "amqp_domain.amqp." ipset ipv4 "amqp_set" no-ttl
    local-zone: "global.net" ipset

ipset:
    name-v4: "test_set"
```

Would allow for an iptables setup as follows, ensuring no overlap between the rules.

```bash
ipset -A OUTPUT -m tcp -p tcp --dport 443 -m set --match-set https_set dst -J ALLOW
ipset -A OUTPUT --dport 5431 -m set --match-set amqp_set dst -J ALLOW
ipset -A OUTPUT -m tcp -p tcp --dport 443 -m set --match-set test_test dst -J ALLOW
```

## Notes

There are a few things to note, obviously this needs to have Unbound build with ipset support via `--enable-ipset` and the apparmor profile modified to include the `CAP_NET_ADMIN` permission as well (necessitated by netlink framework).
These changes would need to be included in the Debian package distribution and also potentially some refactor to allow the apparmor profile to be conditionally modified to include the capability when the USE_IPSET env var is detected from the `configure` script (I will raise a separate issue to deal with this as it is a more general problem to be solved with the Debian package).

In a BSD distribution, compiled with the packet filter framework, there is no support for TTLs to be set on individual table entries. The only support that pf provides is invoking manual expiry of table entries past a delta of `n` seconds via the `pfctl -t <table> -T expire <seconds>` flag (<https://man.openbsd.org/pfctl#T~3>).
Thus no support is added there for TTLs and a suitable warning is raised from the parser when the config is checked.

## Chore Additions

* Added tests for both old and new ipset declaration format
* Updated documentation to deprecate old ipset format and detail new format as well as limitation of no rule TTLs on BSD distributions using packet filter.


ipset/ipset.c
```diff
@@ -170,42 +180,88 @@ static int add_to_ipset(filter_dev dev, const char *setname, const void *ipaddr,
 	mnl_attr_put(nlh, (af == AF_INET ? IPSET_ATTR_IPADDR_IPV4 : IPSET_ATTR_IPADDR_IPV6)
 			| NLA_F_NET_BYTEORDER, (af == AF_INET ? sizeof(struct in_addr) : sizeof(struct in6_addr)), ipaddr);
 	mnl_attr_nest_end(nlh, nested[1]);
+    if (ttl >= 0) {
+        nested[2] = mnl_attr_nest_start(nlh, IPSET_ATTR_TIMEOUT);
+        mnl_attr_put(nlh, NLA_F_NET_BYTEORDER, sizeof(time_t), &ttl);
```
EngineersBox: **[Note]** Seems I have an old version of these changes here, should just be a single u32 packed attribute. I'll update this when I get a chance later today.

ipset/ipset.c
```diff
@@ -252,20 +332,29 @@ ipset_check_zones_for_rrset(struct module_env *env, struct ipset_env *ie,
 		if (p->str[plen - 1] == '.') {
 			plen--;
 		}
-
+        int set_af;
+        if (strncasecmp(p->str2, "ipv4", 4) == 0) {
```
EngineersBox, line 336: **[Note]** Out of scope for this PR, but would be nice to make fixed valued properties (like this one) into enums during the parsing stage of the config. That would require support for arbitrary structural representation of configurations that have many layouts and types though. That's a decent bit of work to do.

PR #1167, opened 2024-10-28
Makefile.in: fix occasional parallel build failures around bison rule

trofi:
Without the change `make -j16 --shuffle` occasinally fails to build as:

    $ make -j16 --shuffle
    ...
    bison -y -d -o util/configparser.c ./util/configparser.y
    ...
    /libtool --tag=CC --mode=compile gcc -I.  -I...-openssl-3.3.2-dev/include -I...-libevent-2.1.12-dev/include -I...-expat-2.6.3-dev/include -DSRCDIR=. -g -O2 -flto -fPIE -pthread  -o configparser.lo -c util/configparser.c
    ...
    util/configparser.c:755:3: error: expected ',' or '}' at end of input
    755 |   YYSYMBOL_server_low_rtt = 626,           /* server_low_rtt  */
        |   ^

The build failure happens due to this `Makefile.in` rule:

    util/configparser.c util/configparser.h:  $(srcdir)/util/configparser.y
        @-if test ! -d util; then $(INSTALL) -d util; fi
        $(YACC) -d -o util/configparser.c $(srcdir)/util/configparser.y

For GNU make that means that each of the targets will attempt the rule execution when the file is missing: one for .c file and another for .h file:

    https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html

The workaround is to only run $(YACC) for .c target and use .c as a pre-requisite for an .h file.

Before the change the build fails about every 10-th run. After the change no build failures after 100 successful builds.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -509,9 +509,14 @@ util/configlexer.c:  $(srcdir)/util/configlexer.lex util/configparser.h
 	fi
 	@if test ! -f $@; then echo "No $@ : need flex and bison to compile from source repository"; exit 1; fi
 
-util/configparser.c util/configparser.h:  $(srcdir)/util/configparser.y
+# Builds both util/configparser.c and util/configparser.h.
+# To avoid double-building we split one target out.
+util/configparser.c:  $(srcdir)/util/configparser.y
 	@-if test ! -d util; then $(INSTALL) -d util; fi
 	$(YACC) -d -o util/configparser.c $(srcdir)/util/configparser.y
+
+util/configparser.h: util/configparser.c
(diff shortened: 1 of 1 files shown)


wcawijngaards (approved):
The fix looks very short and nice to be able to build parallel without failure. Thank you for the contribution!

wcawijngaards:
The same issue is fixed with a similar fix, for NSD, in https://github.com/NLnetLabs/nsd/commit/c2f82f409018d6c8c5757ed14698f25ad4b4f9ac . It had a similar problem, so it could be fixed there too.

PR #1169, opened 2024-10-29
fix: lock-free counters for auth_zone up/down queries

sakateka:
It is simplified version of https://github.com/NLnetLabs/unbound/pull/1160

### DIFF
--- a/daemon/stats.c
+++ b/daemon/stats.c
@@ -325,20 +325,8 @@ server_stats_compile(struct worker* worker, struct ub_stats_info* s, int reset)
 	s->svr.num_query_dnscrypt_replay = 0;
 #endif /* USE_DNSCRYPT */
 	if(worker->env.auth_zones) {
-		if(reset && !worker->env.cfg->stat_cumulative) {
-			lock_rw_wrlock(&worker->env.auth_zones->lock);
-		} else {
-			lock_rw_rdlock(&worker->env.auth_zones->lock);
-		}
-		s->svr.num_query_authzone_up = (long long)worker->env.
-			auth_zones->num_query_up;
-		s->svr.num_query_authzone_down = (long long)worker->env.
-			auth_zones->num_query_down;
-		if(reset && !worker->env.cfg->stat_cumulative) {
-			worker->env.auth_zones->num_query_up = 0;
-			worker->env.auth_zones->num_query_down = 0;
-		}
-		lock_rw_unlock(&worker->env.auth_zones->lock);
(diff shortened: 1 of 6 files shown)


gthess:
Hi @sakateka, sorry for the delayed response.

Removing the lock for authzones in #1160 is not a desired feature because the lock is needed during operation to handle zone updates with XFR for example. So I would like to close that PR if you don't mind :)

I like the idea of moving the authzone counters out of locks but putting them in `worker->stats` is not the correct solution because when libunbound is used(the library not the binary), there is no server worker that has stats and that would eventually lead to a crash. I am talking about accessing counters through `env->stats` from `iterator.c` and `authzone.c`.

If you would like to rework the PR you could put the counters in `mesh_area` and you could follow the paradigm of `mesh_area->ans_expired` for example.

sakateka:
> when libunbound is used(the library not the binary), there is no server worker that has stats

Indeed, I missed it, thank you!
I'll rework the PR.

sakateka:
Hi @gthess, it seems I didn't mess up anything and did everything right this time.
Take a look at this version of the patch, please.

daemon/stats.c
```diff
@@ -325,20 +325,11 @@ server_stats_compile(struct worker* worker, struct ub_stats_info* s, int reset)
...
-		if(reset && !worker->env.cfg->stat_cumulative) {
-			worker->env.auth_zones->num_query_up = 0;
-			worker->env.auth_zones->num_query_down = 0;
-		}
-		lock_rw_unlock(&worker->env.auth_zones->lock);
+		s->svr.num_query_authzone_up = (long long)worker->env.mesh->num_query_authzone_up;
+		s->svr.num_query_authzone_down = (long long)worker->env.mesh->num_query_authzone_down;
+	} else {
+		s->svr.num_query_authzone_up = 0;
+		s->svr.num_query_authzone_down = 0;
```
gthess: Not necessarily needed in this case but for consistency with other `mesh->` values these should be added.
The else is not needed since the worker stats are initialized and reset based on configuration.
```suggestion
		s->svr.num_query_authzone_up += (long long)worker->env.mesh->num_query_authzone_up;
		s->svr.num_query_authzone_down += (long long)worker->env.mesh->num_query_authzone_down;
```

gthess (changes requested):
Looking good!
Just a small change and this can be merged.

daemon/stats.c
sakateka: fixed

gthess (approved):
Thanks! Now counting for auth_zone stats can happen lock free per thread and the values can be aggregated for the stat commands.

PR #1170, opened 2024-10-30
Fix chroot manpage description

### LINKED ISSUE #1164: Docs: Improve manpage about "default" chroot
**Current behavior**

Man page is saying: `By default chroot is enabled and the default is "".`

See my comment here: https://github.com/NLnetLabs/unbound/commit/1d315e11cf267241cbd8a9eb530fd98432ab171d#r148392042

We most likely need to rephrase this sentence, since under some distributions like Ubuntu and Fedora, chroot is actually disabled by default. Making this sentence a contradictory.

melroy89:
Rephrase the man page sentence about what the default chroot is set to, which could also be set to: "" by default. In some distributions like Ubuntu and Fedora this is set to "", which means by default disabled.

So you can't say chroot is enabled by default, since that is not always the case.

Fixes: #1164

### DIFF
--- a/doc/Changelog
+++ b/doc/Changelog
@@ -1,3 +1,6 @@
+30 October 2024: ?
+	- Fix #1164: Fix default chroot manpage sentence.
+
 25 October 2024: Yorgos
 	- Fix #1163: Typos in unbound.conf documentation.
 
--- a/doc/Changelog
+++ b/doc/Changelog
@@ -1,6 +1,3 @@
-30 October 2024: ?
-	- Fix #1164: Fix default chroot manpage sentence.
-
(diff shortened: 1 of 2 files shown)


gthess (changes requested):
Thanks! The sentence does read more clearly.
But I would leave the `doc/Changelog` edit out. We update it ourselves when we merge code. We ofcourse attribute the change there and will use the real name if available or username if not. If you have a preference let me know btw.

melroy89:
Done

gthess (approved):

gthess:
(The online documentation is synced when a new version is released)

PR #1174, opened 2024-11-06
Serve expired cache update fixes

gthess:
This tries to fix certain things related to serve-expired and cache updates:
- If resolution fails, back off and use the expired entry for some time (default 5 seconds)
- If serve-expired is set (but not serve-expired-client-timeout) still try to replace a SERVFAIL answer with an expired reponse
- Exclude explicit RRSIG queries to require validation before being cached

### DIFF
--- a/daemon/worker.c
+++ b/daemon/worker.c
@@ -1845,10 +1845,10 @@ lookup_cache:
 				 * its qname must be that used for cache
 				 * lookup. */
 				if((worker->env.cfg->prefetch &&
-					*worker->env.now >= rep->prefetch_ttl) ||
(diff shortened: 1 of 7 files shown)


wcawijngaards (approved):
Comments elsewhere, but approved. Important to have a limit on work for expired contents that is served, otherwise every incoming query could turn into an upstream request, a request attempt for failing lookups.

wcawijngaards (approved):
The changes look good! With serve expired, the norec_ttl is adjusted to stop repeated attempts, for a short duration, and the additional fix is a catch for additional cases where this is needed.

PR #1189, opened 2024-11-24
Fix the dname_str method to cause conversion errors when the domain name length is 255

wenxuan70:
Fix #1188

PR #1197, opened 2024-12-02
dname_str() fixes

gthess:
- Incorporates the fix from #1189,
- Adds unit tests, and
- Homogenizes the input buffer size.


wcawijngaards (approved):
The code looks good. Use of a definition for the buffer to use is an improvement.

gthess:
Thanks!

PR #1204, opened 2024-12-13
ci: set persist-credentials: false for actions/checkout per zizmor suggestion

maertsen:
I don't think we currently have anything in CI that needs the (readonly) checkout credentials persisted to .git/config during the run.


gthess (approved):
Very good suggestion and the CI ran to completion!

gthess:
@maertsen , should I keep the branch around for future improvements or it can go away?

Issue #1202, opened 2024-12-08
DNS over TLS is extremely slow (~40+ms) because of slow TLS handshake


frebib:
**Describe the bug**
Simply `dig @my.dns.server#853 +tls some.query` consistently takes 40-45ms
Configuring nginx to perform the TLS decryption reduces this to 20-30ms initially, then to zero for subsequent requests (reusing tls session?)

**To reproduce**
1.  Configure unbound for TLS with a few simple options:
```
interface: 2a02:8010:64b4::d45a@853
tls-port: 853
tls-service-key: path/to/private.key
tls-service-pem: path/to/fullchain.cer
tls-ciphers: "ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256"
tls-ciphersuites: "TLS_CHACHA20_POLY1305_SHA256:TLS_AES_256_GCM_SHA384:TLS_AES_128_GCM_SHA256"
```
Note that setting tls-ciphers here seems to make no difference
I'm using ecdsa certificates from letsencrypt here (ecdsa-with-SHA384 / secp384r1 / P-384)

2. Perform the dig
3. Observe the TLS handshake taking forever

**Expected behavior**
Unbound should be able to be (nearly) as fast as nginx

**System:**
 - Unbound version: 1.20.0
 - OS: Vyos 1.5.x (although unbound is running in a `alpinelinux/unbound` container)
 - `unbound -V` output:
 ```
 /etc/unbound # unbound -V
Version 1.20.0

Configure line: --build=x86_64-alpine-linux-musl --host=x86_64-alpine-linux-musl --prefix=/usr --sysconfdir=/etc --mandir=/usr/share/man --localstatedir=/var --with-username=unbound --with-run-dir= --with-pidfile= --with-rootkey-file=/usr/share/dnssec-root/trusted-key.key --with-libevent --with-pthreads --disable-static --disable-rpath --enable-dnstap --with-ssl --without-pythonmodule --with-pyunbound
Linked libs: libevent 2.1.12-stable (it uses epoll), OpenSSL 3.3.2 3 Sep 2024
[...]

Issue #1185, opened 2024-11-15
doq bug


liang-hiwin:
```
unbound -V
Version 1.22.0

Configure line: --enable-subnet --with-libevent --without-pthreads --without-solaris-threads --with-ssl --enable-dnscrypt --with-libhiredis --enable-cachedb --enable-tfo-client --enable-tfo-server --enable-ipset --with-ssl=/usr/local/openssl-3.0.15-quic1 --with-libngtcp2=/usr/local/ngtcp2-1.8.1 LDFLAGS=-Wl,-rpath -Wl,/usr/local/ngtcp2-1.8.1/lib --prefix=/usr/local
Linked libs: libevent 2.1.12-stable (it uses epoll), OpenSSL 3.0.15+quic 3 Sep 2024
Linked modules: dns64 cachedb subnetcache ipset respip validator iterator
DNSCrypt feature available
TCP Fastopen feature available

```
#######################################################
```
./unbound -c /usr/local/etc/unbound/unbound.conf -dd -vvvv
[1731685084] unbound[3135312:0] notice: Start of unbound 1.22.0.
[1731685084] unbound[3135312:0] debug: creating udp4 socket 127.0.0.1 5353
[1731685084] unbound[3135312:0] debug: creating tcp4 socket 127.0.0.1 5353
[1731685084] unbound[3135312:0] debug: creating udp4 socket 0.0.0.0 854 doq
[1731685084] unbound[3135312:0] debug: creating tcp4 socket 0.0.0.0 854 tls
[1731685084] unbound[3135312:0] debug: setup TCP for SSL service
[1731685084] unbound[3135312:0] debug: creating udp4 socket 127.0.0.1 5353
[1731685084] unbound[3135312:0] debug: creating tcp4 socket 127.0.0.1 5353
[1731685084] unbound[3135312:0] debug: creating udp4 socket 0.0.0.0 854 doq
[1731685084] unbound[3135312:0] debug: creating tcp4 socket 0.0.0.0 854 tls
[1731685084] unbound[3135312:0] debug: setup TCP for SSL service
[1731685084] unbound[3135312:0] debug: module config: "subnetcache validator iterator"
[1731685084] unbound[3135312:0] debug: chdir to /usr/local/etc/unbound
[1731685084] unbound[3135312:0] debug: chroot to /usr/local/etc/unbound
[1731685084] unbound[3135312:0] debug: drop user privileges, run as root
[...]

PR #1214, opened 2025-01-10
Use TCP_NODELAY on TLS sockets to speed up the TLS handshake.

### LINKED ISSUE #1045: Low Throughput Issue with unbound DNS over TLS on Ubuntu 22.04
I'd like to report a rather peculiar issue that I encountered approximately a year ago. It seems that the reviewers during the evaluation of my paper artifact also stumbled upon this problem. Therefore, I'm curious to understand the root cause of this unusual issue.

I noticed that when testing unbound DNS over TLS (DoT) throughput on Ubuntu 20.04, the throughput was approximately 1000 connections/second. However, when testing the DoT throughput on Ubuntu 22.04, the throughput was unusually low, averaging only around 20 connections/second.

I utilized the configuration file available at: [config_file](https://github.com/Ji-Peng/eng25519_artifact/blob/main/ENG25519/eng25519/test/unbound.conf). Within the same directory, you can find the relevant certificate files, `ed25519.key` and `ed25519.pem`. In the configuration file, I enabled data statistics using "statistics-interval: 60" to monitor throughput-related data.

The client I used is available at: [DoT Timer](https://github.com/Ji-Peng/eng25519_artifact/tree/main/ENG25519/DoT_timer). Its behavior is straightforward: within a timing loop, it establishes a TLS connection, queries a domain name, and then closes the connection.

Both the client and server communicate within an internal network, and my client implementation is very simple. Thus, I suspect the issue does not stem from the client.

I obtained the latest code directly from the `master` branch of `unbound`.

I apologize for not being able to provide a readily available machine to help replicate the issue, as this was a problem I encountered a year ago.

Another clue I can offer is that I suspect the issue is not caused by the computational overhead of the public key cryptography algorithms within OpenSSL. I experimented with using different OpenSSL Engines to accelerate these algorithms and found that regardless of the ENGINE configuration, the throughput remained strangely low.

My current workaround is to use Ubuntu 20.04 instead of 22.04.

I understand that the information provided may not be sufficient to replicate the problem. If you need further details, please let me know, and I'd be happy to assist in reproducing the issue.

gthess:
Setting TCP_NODELAY on TLS sockets speeds up the TLS handshake.
Also noted on https://docs.openssl.org/3.4/man3/SSL_connect/#notes.

It was observed that during the handshake the server waits before sending more handshake data for the client ACK (Nagle's algorithm), which is delayed because the client waits for more data before ACKing (delayed ACK).

Old versions of OSes don't seem to experience the handshake delay (e.g., Ubuntu 20.04).
Tested with the same compiled versions of Unbound and OpenSSL across OS versions.
Maybe something changed in the kernel (or the kernel configuration) but couldn't pinpoint anything in particular.

Fixes #1045, #1185, #1202.

### DIFF
--- a/services/listen_dnsport.c
+++ b/services/listen_dnsport.c
@@ -703,7 +703,10 @@ create_tcp_accept_sock(struct addrinfo *addr, int v6only, int* noproto,
 {
 	int s = -1;
 	char* err;
-#if defined(SO_REUSEADDR) || defined(SO_REUSEPORT) || defined(IPV6_V6ONLY) || defined(IP_TRANSPARENT) || defined(IP_BINDANY) || defined(IP_FREEBIND) || defined(SO_BINDANY)
+#if defined(SO_REUSEADDR) || defined(SO_REUSEPORT)		\
+	|| defined(IPV6_V6ONLY) || defined(IP_TRANSPARENT)	\
+	|| defined(IP_BINDANY) || defined(IP_FREEBIND)		\
+	|| defined(SO_BINDANY) || defined(TCP_NODELAY)
 	int on = 1;
 #endif
 #ifdef HAVE_SYSTEMD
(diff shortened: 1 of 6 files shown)


wcawijngaards (approved):
Better speed on the handshake may improved DNS lookup latency. The code looks fine, for setting the nodelay option.

gthess:
Thanks!

PR #1218, opened 2025-01-13
Draft: Update path to tls-cert-bundle on reload

pemensik:
We need potentially higher privileges when reading key file, but not for reading cert bundle. Try to update also TLS cert path configured for remote TLS servers on reload command.


pemensik:
Hmm, it seems this approach does not help. Even when openssl calls are done, the result is not changed.

gthess:
Hi Petr, I don't have time to look at this at the moment, but I don't think you can change an SSL_CTX after you used it to create SSL connections. You would need to destroy it and make a new one for new options.

PR #1220, opened 2025-01-14
Add unbound members group access to control key

pemensik:
Recent openssl genrsa does not use umask for generated keys. There is no strong reason why every member of unbound group should be able read server key. But control key would be quite useful to be group readable and to allow control access to whole group. Allowing access to control by group membership, not via sudo.


gthess (approved):
Looks good to me.
The old situation was that the group was getting read access on keys and public certificates implicitly.
The current situation is that the group is not getting any read access on keys.
The new situation will be that the group will gain read access on the control key (the one clients need to connect to the server part- Unbound).

Issue #976, opened 2023-12-07
CNAME in local-data: is not resolved


vzuevsky:
**Describe the bug**
```
server:
  local-zone: "fqdn.com." transparent
  local-data: "bb.fqdn.com.      CNAME vv.my.net."
```

**To reproduce**
```
dig bb.fqdn.com @localhost
bb.fqdn.com.	3600	IN	CNAME	vv.my.net.
```

**Expected behavior**
A record corresponding to vv.my.net.

**System:**
Debian 4.19
Version 1.9.0
linked libs: libevent 2.1.8-stable (it uses epoll), OpenSSL 1.1.1n  15 Mar 2022
linked modules: dns64 python subnetcache respip validator iterator

gthess:
Hi @vzuevsky, as per the documentation CNAME is not supported in local zones (https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-local-zone).

If you would like to serve your own content with CNAME support you could use one of stub, forward or auth zone configuration options. The auth zone documentation specifically mentions the CNAME case (https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#authority-zone-options).

GamerBene19:
> Hi @vzuevsky, as per the documentation CNAME is not supported in local zones (https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#unbound-conf-local-zone).
> 
> If you would like to serve your own content with CNAME support you could use one of stub, forward or auth zone configuration options. The auth zone documentation specifically mentions the CNAME case (https://unbound.docs.nlnetlabs.nl/en/latest/manpages/unbound.conf.html#authority-zone-options).

@gthess, can you (or anyone else) go into more detail (or link to an example perhaps), as to how one would set that up? I am struggling with defining the right auth-zone files, and googling does not bring up anything useful (anymore).

Say I want `muh.foo.de` to resolve to (the IP of) `foo.bar` similar to how pihole (which uses dnsmasq iirc) does it, such that this is the reply for a query to `muh.foo.de`:
```
;; ANSWER SECTION:
muh.foo.de.   0	IN	CNAME	foo.bar.
foo.bar.      0	IN	A	192.168.20.30
```


My understanding is that I would need to have these two records _somewhere_:
```
foo.bar. A 192.168.20.30
muh.foo.de. CNAME foo.bar.
```

[...]

PR #1221, opened 2025-01-14
Consider auth zones when checking for forwarders

gthess:
This punches holes into the forwarders tree, similar to stub zones, for auth zones to be considered during forwarder selection.

It was brought up in #976.


wcawijngaards (approved):
Rounds out the support for auth zones, in this case for forwarder configuration at the same time. The code looks nice too!

gthess:
Thanks!

PR #1222, opened 2025-01-15
Unique DoT and DoH SSL contexts to allow for different ALPN

gthess:
To follow the recent interoperability TLS fixes from NSD.

I would appreciate a second pair of eyes especially for the tls_session_ticket_keys.


daemon/unbound.c
```diff
@@ -463,6 +463,58 @@ detach(void)
...
+			cfg->tls_session_ticket_keys.first->str[0] != 0) {
+			if(!listen_sslctx_setup_ticket_keys(
+				cfg->tls_session_ticket_keys.first)) {
+				fatal_exit("could not set session ticket SSL_CTX");
+			}
+		}
+		(void)setup_listen_sslctx(&daemon->listen_dot_sslctx, 1, 0, cfg);
+		(void)setup_listen_sslctx(&daemon->listen_doh_sslctx, 0, 1, cfg);
+	}
+#ifdef HAVE_NGTCP2
```
wcawijngaards, line 507: This used to be inside the `cfg->ssl_service_key` if statement clause. It checks if the key exists, I think, so likely harmless.

wcawijngaards (approved):
This sets the ALPN right on listening SSL transports.

The ssl ticket keys, are now using the same list in memory and same config for both types of transport. I guess if the ticket key is there it can provide session resumption.

daemon/unbound.c
gthess: You mean the quic listening service? Then you are right, let me fix that.

wcawijngaards: Yes, I meant the quic listening socket call. Probably only affects things when configuration is wrong, it is missing the server certificate.

gthess: Fixed (but here it won't show the change because nothing changed in the shown selection)

gthess:
A couple of small additions for ALPN selection and optional context creation.

PR #1224, opened 2025-01-18
Do not use DSA API unless USE_DSA is set

botovq:
Even if USE_DSA is unset, unbound ends up linking against OpenSSL DSA API because these guards are missing.


gthess:
This looks good to me, disabling the DSA calls when DSA support is not compiled in (default). Thanks!

PR #1241, opened 2025-02-19
Fix infra-keep-probing for low infra-cache-max-rtt values

gthess:
The maximum value of a probe rto was not aligned with the (configurable) `infra-cache-max-rtt` value. That could result in `infra-keep-probing` not working if an `infra-cache-max-rtt` value was chosen that was below 12000 ms.

This fix still uses a default value of 12000 ms for the probe but caps it to the `infra-cache-max-rtt` if that is lower.


wcawijngaards (approved):
For those low values of rtt, the probe continuation is fairly nice to have. The code looks nice, expanded with a unit test.

PR #1259, opened 2025-03-31
Review comments for the fast reload feature

gthess:


wcawijngaards (approved):
Thank you for the changes, the fix for the xfr compare stops an awkward bug, and the new documentation text is a good improvement! The changes look good to me.

PR #1265, opened 2025-04-10
Fix WSAPoll

gthess:


wcawijngaards (approved):
The code looks good. The fixes make it more portable.

PR #1280, opened 2025-05-09
Fix auth nsec3 code

gthess:
Fix NSEC3 code to not break on broken auth zones that include unsigned out of zone (above apex) data. Could lead to hang while trying to prove a wildcard answer. Reported by Dmitrii Kuvaiskii from Amazon Web Services.

### DIFF
--- a/services/authzone.c
+++ b/services/authzone.c
@@ -2413,6 +2413,8 @@ az_find_wildcard(struct auth_zone* z, struct query_info* qinfo,
 	if(!dname_subdomain_c(nm, z->name))
 		return NULL; /* out of zone */
 	while((node=az_find_wildcard_domain(z, nm, nmlen))==NULL) {
+		if(nmlen == z->namelen)
+			return NULL; /* top of zone reached */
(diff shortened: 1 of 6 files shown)


wcawijngaards (approved):
This fixes the issues in the places it was reported and also extra checks in other places. The code looks nice.

dimakuv:
Five minutes late to the party, but I verified this patch on my end, using a clean Amazon Linux 2023 Docker image. It works correctly:
```bash
$ history | awk '{$1=$1;print}' | cut -d' ' -f2-

# without the fix
git clone https://github.com/NLnetLabs/unbound.git
cd unbound/
git checkout bugfix/auth-nsec3-hang-on-broken-zone~4
./configure
make -j
make install
vim testdata/auth_nsec3_wild.rpl # add the malformed line to induce hang
make test
... hangs ...

# with the fix
make distclean
git checkout bugfix/auth-nsec3-hang-on-broken-zone
./configure
make -j
make install
make test
... no hang, success!
```

gthess:
Thanks for the verification!

PR #1289, opened 2025-05-29
Add extra statistic to track the number of signature validation operations

rijswijk:
This PR adds an extra statistic to the extended statistics to track the number of signature validation operations performed by the validator module. This is mostly useful for research purposes, as it makes it possible to track the workload required for cryptographic operations over time.

The statistic tracks the number of calls to `verify_canonrrset(..)` in `val_sigcrypt.c`. The implementation follows the same pattern as for the statistic that tracks the number of bogus RRsets. That is: there is a lock around operations that change or read the statistic (I assume this is because the validator module is not replicated in each thread if Unbound runs multi-threaded).

It would be much appreciated if this patch could be considered for merging as it would help us in research into performance of PQC algorithms; I admit that the extra lock/unlock operation around updates of the statistic adds overhead, I didn't immediately see an easy way to mitigate this as information on whether or not extended statistics are enabled is not available at the abstraction level where the statistic needs to be updated. It seems this is deemed acceptable for tracking the number of bogus RRsets, so hoping one more lock/unlock is fine here.

### DIFF
--- a/validator/val_sigcrypt.c
+++ b/validator/val_sigcrypt.c
@@ -57,6 +57,7 @@
 #include "sldns/sbuffer.h"
 #include "sldns/parseutil.h"
 #include "sldns/wire2str.h"
+#include "services/mesh.h"
 
 #include <ctype.h>
 #if !defined(HAVE_SSL) && !defined(HAVE_NSS) && !defined(HAVE_NETTLE)
@@ -1678,9 +1679,8 @@ dnskey_verify_rrset_sig(struct regional* region, sldns_buffer* buf,
 	sec = verify_canonrrset(buf, (int)sig[2+2],
 		sigblock, sigblock_len, key, keylen, reason);
 
-	lock_basic_lock(&ve->valops_lock);
-	ve->num_val_ops++;
-	lock_basic_unlock(&ve->valops_lock);
+	/* count validation operation */
+	qstate->env->mesh->val_ops++;
 	
 	if(sec == sec_status_secure) {
 		/* check if TTL is too high - reduce if so */
--- a/validator/val_sigcrypt.c
+++ b/validator/val_sigcrypt.c
@@ -1677,6 +1677,10 @@ dnskey_verify_rrset_sig(struct regional* region, sldns_buffer* buf,
 	/* verify */
 	sec = verify_canonrrset(buf, (int)sig[2+2],
 		sigblock, sigblock_len, key, keylen, reason);
+
+	lock_basic_lock(&ve->valops_lock);
+	ve->num_val_ops++;
+	lock_basic_unlock(&ve->valops_lock);
 	
 	if(sec == sec_status_secure) {
 		/* check if TTL is too high - reduce if so */
--- a/validator/val_sigcrypt.c
+++ b/validator/val_sigcrypt.c
@@ -1680,7 +1680,8 @@ dnskey_verify_rrset_sig(struct regional* region, sldns_buffer* buf,
 		sigblock, sigblock_len, key, keylen, reason);
 
 	/* count validation operation */
-	qstate->env->mesh->val_ops++;
+	if(qstate->env->mesh)
+		qstate->env->mesh->val_ops++;
 	
 	if(sec == sec_status_secure) {
 		/* check if TTL is too high - reduce if so */
--- a/validator/val_sigcrypt.c
+++ b/validator/val_sigcrypt.c
@@ -1680,7 +1680,7 @@ dnskey_verify_rrset_sig(struct regional* region, sldns_buffer* buf,
 		sigblock, sigblock_len, key, keylen, reason);
 
 	/* count validation operation */
-	if(qstate->env->mesh)
+	if(qstate && qstate->env && qstate->env->mesh)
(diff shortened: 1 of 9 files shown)


gthess:
Since this is about every single validation operation I would be concerned about the locking.
I would prefer if the statistic would live/behave similar to `ans_bogus` https://github.com/NLnetLabs/unbound/blob/81f3de4da2e4a35d04465471b436f8253719796f/services/mesh.h#L133 where it is stored per thread and later added together when producing the stats.
In `dnskey_verify_rrset_sig()` you can get access to the `mesh_area` with `qstate->env->mesh`.
As for the extended statistics, you can get access to the value via `qstate->env->cfg->stat_extended`.

rijswijk:
> Since this is about every single validation operation I would be concerned about the locking. I would prefer if the statistic would live/behave similar to `ans_bogus`
> 
> https://github.com/NLnetLabs/unbound/blob/81f3de4da2e4a35d04465471b436f8253719796f/services/mesh.h#L133
> 
> where it is stored per thread and later added together when producing the stats.
> In `dnskey_verify_rrset_sig()` you can get access to the `mesh_area` with `qstate->env->mesh`.
> As for the extended statistics, you can get access to the value via `qstate->env->cfg->stat_extended`.

I copied this pattern from how `ans_bogus` is counted, which also involves a lock it seems, I found this pattern here:
https://github.com/NLnetLabs/unbound/blob/e4cf7aeccfa165c509719a4c8da9140f966476e6/validator/val_utils.c#L453

This led me to conclude that it was OK to use locking for this type of statistic, so I'm now somewhat confused as to how `ans_bogus` is actually getting counted?

gthess:
You followed the pattern for `num_rrset_bogus` perfectly :)
(But after this PR I would also like to take the `num_rrset_bogus` stat outside of the validator module)

You followed the pattern for `num_rrset_bogus` (lives in the validator module and has to be locked for sharing with threads) but not the pattern for `ans_bogus` (lives in the mesh area which is per thread, it has to be added together with the rest of the per thread instances when producing the final stats).

(small note that `ans_bogus` is also counted in the worker for cache replies which do not have a mesh area, but I'll skip mentions to that for clarity, maybe :)

For the `ans_bogus` that lives in the mesh area:

It is defined here (for stats):
https://github.com/NLnetLabs/unbound/blob/e4cf7aeccfa165c509719a4c8da9140f966476e6/libunbound/unbound.h#L772
and here (for the mesh area):
https://github.com/NLnetLabs/unbound/blob/e4cf7aeccfa165c509719a4c8da9140f966476e6/services/mesh.h#L133

It is counted (for the mesh area) here:
https://github.com/NLnetLabs/unbound/blob/e4cf7aeccfa165c509719a4c8da9140f966476e6/services/mesh.c#L1450

It is gathered here:
https://github.com/NLnetLabs/unbound/blob/e4cf7aeccfa165c509719a4c8da9140f966476e6/daemon/stats.c#L275

You can follow this paradigm and when counting for `val_ops` in `dnskey_verify_rrset_sig()` you can reference it through `qstate->env->mesh`.

rijswijk:
Ahh, the naming is confusing, you're (of course) right. Let me see if I can find some time to fix this tomorrow.

rijswijk:
With thanks to the rain today I finally found time to make the requested change 😅

validator/val_sigcrypt.c
```diff
@@ -1677,6 +1678,9 @@ dnskey_verify_rrset_sig(struct regional* region, sldns_buffer* buf,
 	/* verify */
 	sec = verify_canonrrset(buf, (int)sig[2+2],
 		sigblock, sigblock_len, key, keylen, reason);
+
+	/* count validation operation */
+	qstate->env->mesh->val_ops++;
```
wcawijngaards: The mesh is a null pointer, in some cases. This fix then does not crash but also does not count that validation operation. That would not count offline ZONEMD validation operations, when the server is started and reads authority zones. I think this is why the ci run is failed.
```suggestion
	if(qstate->env->mesh)
		qstate->env->mesh->val_ops++;
```

rijswijk: Ah good point, I missed that.

rijswijk: Hmm, this doesn't solve it:

```
0x000055555559e22d in dnskey_verify_rrset_sig (region=<optimized out>, buf=<optimized out>, ve=0x7fffffffdaa0, 
    now=<optimized out>, rrset=0x55555582e3c0, dnskey=<optimized out>, dnskey_idx=0, sig_idx=0, sortree=0x7fffffffd828, 
    buf_canon=0x7fffffffd824, reason=0x7fffffffda18, reason_bogus=0x0, section=LDNS_SECTION_ANSWER, qstate=0x0)
    at validator/val_sigcrypt.c:1683
1683		if(qstate->env->mesh)
```

Apparently `qstate` can also be `NULL`, so I guess the code should be something like:

```
if (qstate && qstate->env && qstate->env->mesh) [...]
```

rijswijk: Pushed a fix.

gthess:
This looks good to me now, thanks for the fixes!

PR #1297, opened 2025-06-19
edns-subnet: fix NULL_AFTER_DEREF on subnetmod

bewflast:
Found by static analyzer svace.
Static analyzer message: Redundant comparison with a NULL value at subnetmod.c:236 for pointer 'sn_env->subnet_msg_cache', which was dereferenced at slabhash.c:228 by passing as 1st parameter to function 'slabhash_setmarkdel' at subnetmod.c:235.

Moved usage of sn_env->subnet_msg_cache in slabhash_setmarkdel after checking.


wcawijngaards (approved):
The code patch fixes the analyzer warning. The code deals with alloc failure in a better way, so that is nice.

Issue #1247, opened 2025-03-04
forward-first: ssl handshake failed on root nameservers.


planetf1:
**Describe the bug**
With 'forward-first' active, unbound attempts to connect to root nameservers using SSL

```
<29>1 2025-03-04T08:40:44+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="36"] [1838:3] notice: ssl handshake failed 192.33.4.12 port 53
<27>1 2025-03-04T08:41:03+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="37"] [1838:2] error: ssl handshake failed: channel closed
<29>1 2025-03-04T08:41:03+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="38"] [1838:2] notice: ssl handshake failed 192.112.36.4 port 53
<27>1 2025-03-04T08:41:03+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="39"] [1838:2] error: ssl handshake failed: channel closed
<29>1 2025-03-04T08:41:03+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="40"] [1838:2] notice: ssl handshake failed 2801:1b8:10::b port 53
<27>1 2025-03-04T08:41:04+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="41"] [1838:2] error: ssl handshake failed: channel closed
<29>1 2025-03-04T08:41:04+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="42"] [1838:2] notice: ssl handshake failed 2801:1b8:10::b port 53
<27>1 2025-03-04T08:41:14+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="43"] [1838:3] error: ssl handshake failed: channel closed
<29>1 2025-03-04T08:41:14+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="44"] [1838:3] notice: ssl handshake failed 2001:500:12::d0d port 53
<27>1 2025-03-04T08:41:34+00:00 OPNsense.cherrybyte.me.uk unbound 1838 - [meta sequenceId="45"] [1838:2] error: ssl handshake failed: channel closed
[...]

PR #1298, opened 2025-06-25
dot and forward-first: make sure the recursion fallback does not use TLS (issue #1247)

matthias-lay:
this fixes the issue #1247 by hard resetting ssl in the fallback case, when a DoT Forwarder is used with _forward-first=yes_ set.

As this might impact security, I added a logging statement if this occurs


gthess:
Hi @matthias-lay,

Thanks a lot for this but as commented in https://github.com/NLnetLabs/unbound/issues/1247#issuecomment-3004544904 it is only a partial fix.

I also identified another issue when falling back from the forwarder that is also addressed in the #1247 referenced commits.

matthias-lay:
thx for the info. and great to see this fixed

PR #1299, opened 2025-07-01
Fix typos

jlduran:


wcawijngaards (approved):
The spelling fixes are good to have! The fixes in generated content, are really fixes for the content from the programs libtoolize and aclocal. The spelling fixes are going to disappear as the files are generated again, they can be fixed in the original place. And then would, eventually with updates, and generated again, appear here.

jlduran:
Thank you! Yes, libtool has been patched upstream, and yes, it needs a regen.

PR #1328, opened 2025-09-03
fixing missing dependencies in Makefile.in

Meiye-lj:
This PR fixes an issue in the Makefile. Specifically, previously, any modifications of files like validator/val_nsec3.h would not trigger a rebuild of stats.o. The PR fixes this by including them as additional dependencies.


wcawijngaards:
The make depend, that was run, has produced an output that includes the additional dependencies from the change requested. That means those file dependencies are updated, and it uses the regular dependency method for it, that is preferable. The change is not needed with that fix, hence the pull request is closed without a merge, the commit fixes it in a different way. Thank you for noting the missing dependencies!

PR #1331, opened 2025-09-08
Replace deprecated $function by new $action

jplesnik:
The long-deprecated $function was removed from future SWIG 4.4.0. It can be safely replaced by $action.

More information about swig change could be found in the commit [Remove long-deprecated $function variable](https://github.com/swig/swig/commit/3be7697a33c98435865fc713f8ecf7cf10765f13)


gthess (approved):
Based on the [online docs](https://swig.org/Doc1.3/SWIGDocumentation.html#exception), `$action` seems to be there since 1.3 and `$function` barely mentioned back then.

Thanks for this!

PR #1334, opened 2025-09-12
TSIG for Zone Transfers

wcawijngaards:
This adds TSIG for zone transfers. For auth-zones and RPZ zones, the AXFR and IXFR can be with TSIG.

TSIG keys are configured as a `tsig-key:` entry with `name: <key name>` and `algorithm: <algorithm name>` and `secret: <base64 blob>`.

The `auth-zone:` can have `primary-tsig: <addr or name> <TSIG key name>` and `allow-notify-tsig: <addr or name or netblock> <TSIG key name>`. Also an `rpz:` can have `primary-tsig` and `allow-notify-tsig`. If the notify is set to use TSIG, then unbound has to receive the notify signed with that TSIG key to process it, it is refused otherwise. The primary is contacted, for SOA probe and for AXFR and for IXFR, with the TSIG for the zone transfer.

This conforms with RFC2845 and RFC8945. The unit test has the digests match TSIG digests from NSD and dig.


phirince:
Any plans to get this PR merged ?

-PP

gthess:
Hi @phirince, it is on our roadmap but big feature changes are on pause at the moment since we are focusing more on security reports as you can see from our recent releases.

PR #1335, opened 2025-09-13
[Feat] Unbound: Support per-rule (local-zone) ipset targets

EngineersBox:
> NOTE: This PR was previously open here (https://github.com/NLnetLabs/unbound/pull/1162),
> but was closed to fix the branch name, and is thus re-opened here.

This PR addresses the issue: <https://github.com/NLnetLabs/unbound/issues/1342>

# Notes

There are a few things to note, obviously this needs to have Unbound build with ipset support via `--enable-ipset` and the apparmor profile modified to include the `CAP_NET_ADMIN` permission as well (necessitated by netlink framework).
These changes would need to be included in the Debian package distribution and also potentially some refactor to allow the apparmor profile to be conditionally modified to include the capability when the USE_IPSET env var is detected from the `configure` script (I will raise a separate issue to deal with this as it is a more general problem to be solved with the Debian package).

In a BSD distribution, compiled with the packet filter framework, there is no support for TTLs to be set on individual table entries. The only support that pf provides is invoking manual expiry of table entries past a delta of `n` seconds via the `pfctl -t <table> -T expire <seconds>` flag (<https://man.openbsd.org/pfctl#T~3>).
Thus no support is added there for TTLs and a suitable warning is raised from the parser when the config is checked.

# Chore Additions

* Added tests for both old and new ipset declaration format
* Updated documentation to deprecate old ipset format and detail new format as well as limitation of no rule TTLs on BSD distributions using packet filter.


EngineersBox:
Currently, the DNS name to ipset mappings are stored in a list which is linear searched. This is obviously sub-optimal, however I am unsure as to what the preferred approach would be for this repo. I noted a RB-tree implementation that could serve as an efficient way to maintain the mappings and avoid the linear search cost. Happy to discuss and plan changes as necessary.

PR #1337, opened 2025-09-15
0 TTL cached replies and some TTL behavior changes

gthess:
This is built on top of #1181 but instead tries to treat cached records that reach TTL 0 as expired.

It also does the following since I have been poking around with TTL behavior:
- 0 TTL upstream answers are no longer cached by cachedb, as they should
- 'serve-expired-reply-ttl' is changed and is now capped by the original
  TTL value of the record to try and make some sense when replying
  with expired records.
- TTL decoding was updated to adhere to RFC8767 section 4 where a set
  high-order bit means the value is positive instead of 0.


util/data/msgencode.c
```diff
@@ -496,10 +496,18 @@ packed_rrset_encode(struct ub_packed_rrset_key* key, sldns_buffer* pkt,
 				return r;
 			sldns_buffer_write(pkt, &key->rk.type, 2);
 			sldns_buffer_write(pkt, &key->rk.rrset_class, 2);
-			if(data->rr_ttl[j] < adjust)
+			if(key->rk.flags & PACKED_RRSET_UPSTREAM_0TTL) {
+				sldns_buffer_write_u32(pkt, 0);
+			} else if(adjust == 0) {
+				sldns_buffer_write_u32(pkt, data->rr_ttl[i]);
```
wcawijngaards: The rrset rotation moved to `j`, but the code uses index `i`.
```suggestion
				sldns_buffer_write_u32(pkt, data->rr_ttl[j]);
```

util/data/msgencode.c
```diff
@@ -496,10 +496,18 @@ packed_rrset_encode(struct ub_packed_rrset_key* key, sldns_buffer* pkt,
...
+			} else if(TTL_IS_EXPIRED(data->rr_ttl[j], adjust)) {
 				sldns_buffer_write_u32(pkt,
-					SERVE_EXPIRED?SERVE_EXPIRED_REPLY_TTL:0);
-			else	sldns_buffer_write_u32(pkt, data->rr_ttl[j]-adjust);
+					EXPIRED_REPLY_TTL_CALC(
+					    data->rr_ttl[i], data->ttl_add));
```
wcawijngaards: We are inserting rr j.
```suggestion
					    data->rr_ttl[j], data->ttl_add));
```

util/data/msgencode.c
```diff
@@ -496,10 +496,18 @@ packed_rrset_encode(struct ub_packed_rrset_key* key, sldns_buffer* pkt,
...
+			} else {
+				sldns_buffer_write_u32(pkt,
+					data->rr_ttl[i] - adjust);
```
wcawijngaards: We are inserting rr j.
```suggestion
					data->rr_ttl[j] - adjust);
```

validator/val_neg.c
```diff
@@ -1066,11 +1066,7 @@ grab_nsec(struct rrset_cache* rrset_cache, uint8_t* qname, size_t qname_len,
 		qname, qname_len, qtype, qclass, flags, now, 0);
 	struct packed_rrset_data* d;
 	if(!k) return NULL;
-	d = (struct packed_rrset_data*)k->entry.data;
-	if(d->ttl < now) {
```
wcawijngaards, line 1070: Should this have `if(TTL_IS_EXPIRED(d->ttl, now))` ?

wcawijngaards (approved):
This makes a bunch of TTL changes. The code looks very good, and 0 time left cache entries are considered not usable.

validator/val_neg.c
gthess: The ttl is already checked in the `rrset_cache_lookup` call above, it would already return NULL if it failed.

wcawijngaards: That works out, I guess that means the expired ttl does not have to get checked twice.

gthess:
Thanks for the review and the index catch!
I'll keep this here and merge back after the 1.24.0 release.

PR #1350, opened 2025-09-23
unbound.service.in: allow CAP_NET_ADMIN

Maryse47:
Allowing `CAP_NET_ADMIN` is necessary for `SO_SNDBUFFORCE` and `SO_RCVBUFFORCE` calls.

See https://man7.org/linux/man-pages/man7/socket.7.html


gthess:
I believe we can also remove `CAP_NET_RAW` if `CAP_NET_ADMIN` is re-introduced. We were using `CAP_NET_RAW`  for `IP_TRANSPARENT` which is also available with `CAP_NET_ADMIN`.

Maryse47:
Removed CAP_NET_RAW

gthess (approved):

gthess:
Thanks!

PR #1351, opened 2025-09-23
ac_cv_func_malloc_0_nonnull for malloc(0) check

gthess:
Patch from Helmut Grohne.


wcawijngaards (approved):
The changes look fine to have. Would need to manage version and synchronisation for the m4 file, for the update.

PR #1352, opened 2025-09-23
pythonmod: fix HANDLE_LEAK on pythonmod_init

petrvaganoff:

Handle 'script_py' is created at pythonmod.c:436
by calling function 'fopen' and lost at pythonmod.c:457,465.


gthess (approved):

gthess:
Thanks! This solves the leak warning when shutting down.

PR #1361, opened 2025-10-15
Fix for analysis and ports workflows iOS, Windows

wcawijngaards:
Fix the analysis and ports workflow.

- Remove iPhone armv7s, and iPhoneSimulator i386 from ios ci.
  The lib system does not provide symbols for it on the new macos
  runner.
- Fix to exclude libssp for windows compiles.


gthess (approved):
Looks good to me!
Nice that we have the analysis ci back on track!

PR #1365, opened 2025-10-24
Use $action instead of $function in python SWIG interface

pemensik:
$function is not supported since SWIG 4.4.0.

Fedora Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2405293


pemensik:
Duplicate, closing in favor [1331](https://github.com/NLnetLabs/unbound/pull/1331)

PR #1374, opened 2025-11-13
Mesh reply counters

edmonds:
This branch adds two new statistics counters:

* `num.queries.replyaddr_limit`: This counts the number of queries dropped due to encountering the limit on the number of mesh reply addresses queued. This allows distinguishing drops due to this limit, separately from other kinds of drops. (E.g., see https://github.com/NLnetLabs/unbound/pull/1159 which also removed several other kinds of drops from being tracked by `mesh_area::stats_dropped` and introduced distinct counters for those kinds of drops.)
* `requestlist.current.replies`: This counts the current number of mesh reply entries that have been queued. Basically it exports `mesh_area::num_reply_addrs`. This allows distinguishing the number of request list entries (`requestlist.current.user`) from the number of reply addresses queued (`requestlist.current.replies`) which are waiting on those request list entries. The latter can legitimately be very high even if the former is low, if there are a lot of clients waiting to resolve the same query.

Thanks!

### DIFF
--- a/daemon/remote.c
+++ b/daemon/remote.c
@@ -847,6 +847,8 @@ print_stats(RES* ssl, const char* nm, struct ub_stats_info* s)
 		(unsigned long)s->mesh_num_states)) return 0;
 	if(!ssl_printf(ssl, "%s.requestlist.current.user"SQ"%lu\n", nm,
 		(unsigned long)s->mesh_num_reply_states)) return 0;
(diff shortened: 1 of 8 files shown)


wcawijngaards (approved):
The code looks nice and fits with the existing code. The extra counters are useful to disambiguate what is going on when there are drops.

PR #1381, opened 2025-11-24
Do not initialize quic_table unless it is enabled

pemensik:
Fedora in FIPS mode might fail to initialize ngtcp2 library, because some ciphers desired are not available.

Make it possible to skip initialization by setting explicitly quic_port to 0. Unless we have some listeners for port 853 configured, skip its initialization as well.

Related: https://pagure.io/freeipa/issue/9877

### DIFF
--- a/daemon/daemon.c
+++ b/daemon/daemon.c
@@ -558,9 +558,11 @@ daemon_create_workers(struct daemon* daemon)
 	verbose(VERB_ALGO, "total of %d outgoing ports available", numport);
 
 #ifdef HAVE_NGTCP2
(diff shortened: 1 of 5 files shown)


pemensik:
I have made it possible to disable QUIC usage by: ``quic-port: 0``. I think it were not possible to serve only TLS on port 853 when QUIC support was built-in. I thought initially there is no better config check. But then found there is already function to check quic is used.

wcawijngaards (approved):
The non-initialization of the quic structure when it is not used, seems like a good idea. The code looks fine.

PR #1383, opened 2025-11-28
- Fix randomness generation on macOS/iOS under chroot

jdek:
SecRandomCopyBytes() has existed since macOS 10.7 (2011) and iOS 2.0 (2008), and is the primary API for cryptographic random numbers.


jdek:
Ping @wcawijngaards; would be really nice to get unbound working under chroots, also this is a massive simplification of hand-rolled crypto to system primitives.

wcawijngaards:
So that is certainly nice to have it work under chroot. Currently the header still says it is a copy of `lib/libcrypto/arc4random/getentropy_osx.c` in OpenBSD, and it would be nicer to update that so it does not confuse, I think adding a comment that says basically the issue explanation. It means not supporting old OSX versions.

jdek:
Thanks! And good catch on the documentation note.

PR #1385, opened 2025-12-14
Change unbound.service to not duplicate if chroot and run are equal.

ninetailedtori:
Not sure if I missed documentation at all or needed to do anything else. Lemme know, as it's my first time contributing to unbound :]


ninetailedtori:
Well, it'll be a little convoluted if https://github.com/NLnetLabs/unbound/pull/1387 passes.

PR #1394, opened 2026-01-21
Log Linux thread ID

gthess:
Adds the ability to log the system-wide Linux thread ID instead of Unbound's internal thread counter.
It can help when debugging with system tools.

### DIFF
--- a/doc/Changelog
+++ b/doc/Changelog
@@ -1,3 +1,9 @@
+1 November 2025: Yorgos
+	- Fix #1366: Infra cache does not work correctly for NAT64, by
+	  moving the NAT64 synthesis from the iterator when selecting a target
+	  address, to the delegation point itself when adding target
+	  addresses.
+
 27 October 2025: Yorgos
 	- Merge #1331 from Jitka Plesníková: Replace deprecated $function by
 	  new $action, for SWIG.
--- a/iterator/iter_utils.c
+++ b/iterator/iter_utils.c
@@ -253,7 +253,9 @@ iter_apply_cfg(struct iter_env* iter_env, struct config_file* cfg)
 	return 1;
 }
 
-/** filter out unsuitable targets
+/** filter out unsuitable targets.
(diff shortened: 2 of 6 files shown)


wcawijngaards (approved):
The change looks okay. The code change from daemon/worker.c, in worker_init, can move to the thread_start routine, where the log_thread_set used to be. And bring along the gettid() call, already initialising a piece of the worker structure. Then the log information is set before possible errors printed by later ub_thread_blocksigs calls, or pipe close calls. Otherwise those have thread 0 because not set, in the logs. That is acceptable too, the errors are not expected there. The move can be nicer, and only because of error printout in util/tube.c and util/locks.c. Approved either way, for the change.

gthess:
Thanks for the review.
I see what you mean, but then I would have to put the `gettid()` call in 2 places.
Once in `thread_start()` for all the other workers, and once in `daemon_fork()` after the call to `daemon_create_workers` probably. Let me see and run some tests.

edmonds:
@gthess @wcawijngaards It would be nice to use `pthread_setname_np()` or `pthread_set_name_np()` (on platforms that have one of those functions, which is more than just Linux) to expose Unbound's internal thread number, for instance by having the worker threads set their thread name to something like `unbound/%u` (with the %u formatted to Unbound's thread number). Many tools can be configured to show the thread name in addition to or instead of the process name (e.g. `top -H`).

gthess:
That's a nice suggestion @edmonds , I am thinking to always do that on systems that support it (no configuration option).

gthess:
@edmonds, that was a deep dive into highly unportable code :)
There is now https://github.com/NLnetLabs/unbound/pull/1400.
Would you be able to test it on platforms available to you that provide such functionality?

edmonds:
I can only test on Linux but I would assume you tested that one already :-)

gthess:
Ha, yes.
I only called you out because you mentioned "on platforms that have one of those functions, which is more than just Linux" and I thought you had such machines ready for testing/use of the feature :)

PR #1398, opened 2026-01-29
RPZ load performance improvements

wcawijngaards:
This adds RPZ zone loading performance improvements. They are build on top of the simdzone parsing changes. The change merges into that branch, so that the RPZ specific changes can be reviewed here.

The dname comparison routines are optimised a bit. The error handling does not convert to string or have the entire RR uncompressed when not needed, but only after the error. The insertion of the localzone data for the RPZ does not perform a tree lookup twice, but once.


wcawijngaards:
A change without the changes from the main branch, is in https://github.com/NLnetLabs/unbound/pull/1399 . That includes only the rpz-zone-load changes visible in the page.

PR #1401, opened 2026-02-09
Add a new build-time option for system TLS

pemensik:
We want to use crypto-policy provided configuration always in our builds. Allow changing the default of tls-use-system-policy-versions at build time by a simple configure parameter.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2080,6 +2080,17 @@ case "$enable_ipsecmod" in
 		;;
 esac
 
+# check for system TLS preference if requested
+AC_ARG_ENABLE(system-tls, AS_HELP_STRING([--enable-system-tls],[Enable preference of system configured TLS socket options]))
+case "$enable_system_tls" in
+	yes)
+		AC_DEFINE([USE_SYSTEM_TLS], [1], [Define to 1 to prefer TLS crypto settings from the system.])
+		;;
+	no|*)
+		# nothing
+		;;
+esac
+
 # check for ipset if requested
 AC_ARG_ENABLE(ipset, AS_HELP_STRING([--enable-ipset],[enable ipset module]))
 case "$enable_ipset" in
--- a/util/config_file.c
+++ b/util/config_file.c
@@ -71,6 +71,9 @@
 #ifdef HAVE_PWD_H
 #include <pwd.h>
 #endif
+#ifndef USE_SYSTEM_TLS
+#define USE_SYSTEM_TLS 0
+#endif
 
 /** from cfg username, after daemonize setup performed */
 uid_t cfg_uid = (uid_t)-1;
@@ -129,7 +132,7 @@ config_create(void)
 	cfg->tls_cert_bundle = NULL;
 	cfg->tls_win_cert = 0;
 	cfg->tls_use_sni = 1;
-	cfg->tls_use_system_policy_versions = 0;
+	cfg->tls_use_system_policy_versions = USE_SYSTEM_TLS;
 	cfg->https_port = UNBOUND_DNS_OVER_HTTPS_PORT;
 	if(!(cfg->http_endpoint = strdup("/dns-query"))) goto error_exit;
 	cfg->http_max_streams = 100;


wcawijngaards (approved):
This configure flag works to change the default setting. The change is fine.

pemensik:
Thanks, that was quick! I did not include ``autoreconf`` changes with this commit intentionally. Of course it will make some changes, but I expect that is needed only once before the release.

gthess:
@pemensik , why is this preferable over a default configuration file?

pemensik:
because we want that default even in unbound-libs linked programs. Changes in configuration files are not applied on modified user config files by default. We have prepared update to unbound 1.24.2, but this would make a regression for any people with modified config file. Our defaults do not propagate automatically, because users should be able to override them. Change of default will make it behave similar to previous state. We want to use crypto-policy openssl settings unless it is explicitly overriden anyway.

pemensik:
Also, even with tools like unbound-host and unbound-anchor, you might sometime need to use TLS configuration from the unbound.conf style snippet with client information only. But then you would have to include also ``tls-use-system-policy-versions: yes`` in all snippets containing only destination IP, server name and TLS certificates.

At the same time that new option would not be accepted by older unbound versions. Changing the default to match our desired policy makes it more user friendly. Once we want to disable TLS 1.2, we will want to do it by changing policies in crypto-policy package anyway.

gthess:
Altering the default for library usage (both linking to third party applications and the Unbound tools) makes sense since normally you don't provide a configuration file for those.
Then the configure value above can specifically target the library context creation.

For Unbound itself however, it makes more sense to provide a default configuration file, rather than the user being surprised when the same configuration file makes Unbound behave differently between a package and source compilation.

pemensik:
In that case I think all we want is to modify manual page and make there note we have changed that default. ``unbound -V`` will show change of the default, that is a reason I made extra build option, not only patch changing value in config_file.c.

Frankly, why is disabling TLS 1.2 considered a good default from your side? Why is disabling that support not done by the config change only, as you propose? Why do you assume disabling it by default is a good idea already? All proper clients will negotiate TLS 1.3 if they support it. If they don't, then they need explicit configuration change to keep having encrypted resolution. Has that decision been backed by any data?

gthess:
I am not arguing about the changed value/behavior.
Rather that apart from the usual default options that someone would expect from a package (install directories, user, chroot, etc), a hard coded changed default value would come as a surprise.
I would expect to see such changes in the default shipped configuration file together with comments why the packager has deviated from the default value, same as it is done with other options currently.

I mean at the end you only change the default value and the user can override it if desired.
But knowing that the default value provided is different between OSes/packages is confusing.

pemensik:
I think unbound should use system defaults if they are managed by the system somehow. That is what we expect on Fedora and RHEL too. I think the best package in this regard would be to use OS default. In this regard I think explicitly disabling TLS 1.2 in the code, but without any uncommented config part, is unfortunate choice you have made. We want TLS setting to use system provided case and would it recommend to you also. If you do not want, then we will override it on our distributions.

But yes, it might be accompanied by a better documentation that we have some downstream differences. We usually want them minimal or none, but if this default is proposed by you, then we want to make a change. Even FIPS or FUTURE crypto policy allows TLS 1.2 in Fedora 43. It might be too early to disable it by default.

gthess:
How about we:
- Change the currently unreleased `tls-use-system-policy-versions` option to `tls-use-system-policy`. Still keeping it default off.
- When that option is on AND `/etc/crypto-policies/config` exists, Unbound would not do anything with versions and ciphers, (but ciphers could still be overridden by the current user configuration options).
- Then a `config`ure option, like the one proposed here, could turn on the `tls-use-system-policy` option for the created library contexts.

So in your case you would be free to bypass the defaults by shipping the `tls-use-system-policy` option and `config`ure to also set it by default for the library.

gthess:
I believe then that your preference would be to have the option `tls-use-system-policy` enabled by default since it would only take effect if `/etc/crypto-policies/config` is in place?

pemensik:
Yes, we prefer always to follow crypto-policy defaults, no config file change needed for it. It should be possible to enable even TLS 1.1 clients by using LEGACY policy. Created [a test](https://gitlab.com/redhat/centos-stream/tests/unbound/-/merge_requests/21), verified that even older version never allowed TLS < 1.2 clients even in LEGACY policy. Yes, when crypto-policy is present, it should not try to change system provided defaults unless administrator told it so.

I think unbound lacks a nice configuration option for TLS versions. Something like [bind9 tls protocols](https://bind9.readthedocs.io/en/v9.20.18/reference.html#namedconf-statement-protocols) statement. It allows configuration of specific ciphers, which should be done only by educated people. But specification which TLS versions are accepted, that should be better understood. But I have not failed a way to specify exiplicitly: I want TLS 1.2+1.3 and nothing else.

Something like ``tls-allow-versions: tls1.2 tls1.3``. Nothing similar is possible now, right?

gthess:
The plan for now seems to be:
- remove this merged configure option,
- remove the currently unreleased `tls-use-system-policy-versions` option. In hindsight it seems like a bazooka to a simple request.
- introduce a `tls-version-minimum` or `tls-version-allow` option only for TLSv1_2 and TLSv1_3. Versions lower than TLSv1_2 would still remain disabled as they are deprecated.

TLSv1_2 would be allowed by default then which solves this issue as well. Users would have the option to turn it off when needed.

I'll probably pick this up soon-ish.

pemensik:
I seems to me hardcoded default is used as a replacement for runtime configuration of accepted versions. Even [nginx](https://nginx.org/en/docs/http/configuring_https_servers.html) has ssl_protocols option. Peeked into lib/isc/tls.c of bind 9.20.18, it uses the same flags as you do, SSL_OP_NO_TLSv1_2 and SSL_OP_NO_TLSv1_3 . But configurable by file statements.

pemensik:
We would like to keep disabling of TLS 1.1 driven by our crypto policy. It allows accepting of TLS 1.1 clients when switching to LEGACY policy. But I admit that is probably meant more for HTTPS clients where some older implementations do exist. I somehow doubt many DNS over TLS clients exist, which do not support at least TLS 1.2. There are many old implementations of DNS, but they typically do not use TLS in that case. Mikrotik for example knows only DNS over HTTPS, but that is served by unbound as well. Not sure which TLS version it uses.

I expect there is no plan now to configure different TLS version for DoH and DoT server sockets, right?

gthess:
Indeed, no individual configuration for this work now.
For older versions than 1.2 you would still need to patch support for them if you like, they will be hardcoded out.

pemensik:
Asked our crypto folks. They said possible deep packet filtering might be done somewhere, where it might reset TLS 1.3 connection. While setting minimal version would be ok most of time, ability to specify only version 1.2 might be useful in some corner cases. I am not sure whether similar problem were not reported in firewall of China for example.

Since I did not notice older TLS versions are always disabled even when chosen crypto policy should allow it, it would be no big deal. But in general those default should be driven by crypto policy, which handle also those cases.

gthess:
So it seems like a "choose which available versions to enable" is the most versatile option, I'll use that then.

gthess:
@pemensik, there is now #1406 with the new `tls-protocols` configuration option.

Issue #1346, opened 2025-09-22
[FR] Please allow back TLS 1.2


erwdavid:
In 1.24 TLS 1.2 was disabled. I have android devices not maintained anymore, that cannot do TLS 1.3 for the "private DNS" feature, would it be possible to at least have an option to re-enable it when we have this type of clients ?

gthess:
The commit above introduces a new option `tls-use-system-policy-versions` with default `no`.
Enabling the option will allow openssl to use the default allowed versions, which may or may not be further filter by a system-wide policy.

gthess:
The fix above that introduced the (unreleased) `tls-use-system-policy-versions` option is reverted in favor of #1406 and the new configuration option `tls-protocols`. TLSv1.2 will be enabled by default.

Issue #1303, opened 2025-07-17
[FR] Disable TLSv1.2


andreasschulze:
**Current behavior**
in unbound.conf, I can configure TLS1.2 and TLSv1.3 ciphers. But I can't configure, which TLS versions must be used.

**Describe the desired feature**
In some cases, if client and server side both support TLSv1.3, it may be usefull to explicit disable TLSv1.2. This is currently not possible. I also _think_, there are documents / drafts in the IETF, saying "TLS1.3 is the preferred protocol" but cant find a reference.

**Potential use-case**
I personally think about a private patch (I build unbound from source) in the spitit of this code: https://github.com/NLnetLabs/unbound/blob/master/util/net_help.c#L1254-L1261 but repeated for "we have TLS1.3, so disable TLS1.2"

Before writing such private patch, I like to ask if this makes sense at all

wcawijngaards:
I have implemented the suggestion. I also believe that TLSv1.3 is deployed fairly widely so that this patch makes sense. Thanks for the heads up!

andreasschulze:
nice :-)

andreasschulze:
NSD has similar code: https://github.com/NLnetLabs/nsd/blob/master/server.c#L2131-L2148

wcawijngaards:
The similar issue for NSD is fixed with https://github.com/NLnetLabs/nsd/commit/e2aaa7c9abca26051906ba571f98dd294928d2fd . OpenSSL 1.1 from 2018 has TLSv1.3 support, so I think this is not likely to be an issue. Until people want to provide DoT to legacy clients, but that also seems unlikely, then it would need an option for allowing the older. For use of nsd-control, and unbound-control, those use the same openssl as the server, so have the same availability of TLS versions, so that works regardless.

gthess:
@andreasschulze, see #1406. TLSv1.2 is again enabled by default but you could explicitly turn it off with the new `tls-protocols` option in that PR.

PR #1406, opened 2026-02-16
Introduce new 'tls-protocols' configuration option

### LINKED ISSUE #1432: DoT server rejects TLS 1.2 clients on FreeBSD: SSL_OP_NO_TLSv1_2 hardcoded in listen_sslctx_setup()
**Describe the bug**
In Unbound 1.24.2 built on FreeBSD 16.0 with OpenSSL 3.5.3, the DoT server (port 853) rejects TLS 1.2 client connections with "Alert Fatal: Protocol Version" (SSL alert number 70).

The root cause is in util/net_help.c inside listen_sslctx_setup(), where SSL_CTX_set_options(ctx, SSL_OP_NO_TLSv1_2) is called unconditionally when SSL_OP_NO_TLSv1_3 is defined in OpenSSL headers. This disables TLS 1.2 on the server context at startup, before any client connects, effectively breaking TLS negotiation for all TLS 1.2 clients.

The SSL_OP_NO_TLSv1_2 block appears to be independent from the /etc/crypto-policies/config check, which only controls the hardcoded cipher list. This means the bug may affect all platforms, not just FreeBSD.

This issue is reproducible on any FreeBSD system running Unbound 1.24.2 with OpenSSL 3.5.3, not specific to pfSense.

Notable affected clients: Windows 11 Dnscache, which still uses the deprecated SCHANNEL_CRED structure and therefore sends TLS 1.2-only Client Hello for DoT connections, making DNS-over-TLS completely non-functional on Windows 11.

**To reproduce**
1. Build Unbound 1.24.2 on FreeBSD 16.0 with OpenSSL 3.5.3
2. Enable DoT server (tls-port: 853) in unbound.conf
3. Connect a TLS 1.2 client to port 853:
   `openssl s_client -connect 127.0.0.1:853 -tls1_2 -servername <hostname>`
4. Observe Alert Fatal: Protocol Version (SSL alert number 70)
5. Verify TLS 1.3 works correctly:
   `openssl s_client -connect 127.0.0.1:853 -tls1_3 -servername <hostname>`

**Expected behavior**
The DoT server should accept both TLS 1.2 and TLS 1.3 clients, negotiating the highest version mutually supported — as TLS protocol was designed to do. Disabling TLS 1.2 on a server breaks compatibility with a large number of legitimate clients.

**Additional context**
Unbound version: 1.24.2
FreeBSD version: 16.0
OpenSSL version: 3.5.3

Configure line:
`--with-ssl=/usr --enable-dnscrypt --with-libevent --with-pthreads --build=amd64-portbld-freebsd16.0`

TLS 1.2 client connection attempt result:
`SSL alert number 70 - Alert Fatal: Protocol Version`

TLS 1.3 client connection result:
`New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384 - handshake successful`

Relevant code in util/net_help.c:

```c
#if defined(SSL_OP_NO_TLSv1_2) && defined(SSL_OP_NO_TLSv1_3)
    /* if we have tls 1.3 disable 1.2 */
    if((SSL_CTX_set_options(ctx, SSL_OP_NO_TLSv1_2) & SSL_OP_NO_TLSv1_2)
        != SSL_OP_NO_TLSv1_2){
        log_crypto_err("could not set SSL_OP_NO_TLSv1_2");
        return 0;
    }
#endif
```

Suggested fix: remove or make this block configurable via unbound.conf (e.g. tls-min-version directive).

gthess:
It specifies which of the supported TLS protocols will be used.
  This change invalidates some previous changes:
  - TLSv1.2 is again enabled by default, but can be selectively turned off if desired (related to #1303).
  - The briefly introduced (not yet released) 'tls-use-system-versions' configuration option, that addressed #1346, is reverted in favor of 'tls-protocols'.
  - The briefly introduced (not yet released) '--enable-system-tls' configure option, related to #1401, is no longer needed with the new option and the current default. 

Fixes #1432.


util/config_file.c
```diff
@@ -132,7 +129,7 @@ config_create(void)
 	cfg->tls_cert_bundle = NULL;
 	cfg->tls_win_cert = 0;
 	cfg->tls_use_sni = 1;
-	cfg->tls_use_system_policy_versions = USE_SYSTEM_TLS;
+	cfg->tls_protocols = strdup("TLSv1.2 TLSv1.3");
```
wcawijngaards: This checks if there is an out of memory condition.
```suggestion
	if(!(cfg->tls_protocols = strdup("TLSv1.2 TLSv1.3"))) goto error_exit;
```

wcawijngaards (approved):
The code looks fine for the tls-protocols selection feature. If it can be used instead of the two other options, that looks to be a neater configuration set up.

PR #1408, opened 2026-02-20
Fix shared memory stats with threads

### LINKED ISSUE #152: Shared memory statistics are not reliable
`unbound` statistics exposed via shared memory are changing unexpectedly even if server itself does no work.

1. I had started a separate `unbound` server with the shm stats enabled:

```
server:
        verbosity: 1
        statistics-interval: 3

        shm-enable: yes
        shm-key: 11777

        statistics-cumulative: yes
        extended-statistics: yes

        num-threads: 3
```

2. Sent three requests to it:

```
$ dig @localhost example.com
 .. skipped a bunch of text for clarity
;; SERVER: ::1#53(::1)
 ..

$ dig @localhost example.org
 .. skipped
;; SERVER: ::1#53(::1)

$ @localhost example.net
 ..
;; SERVER: ::1#53(::1)
```

3. `unbound-control stats_noreset` shows that three IPv6 queries were issued:
```
$ unbound-control stats_noreset | grep num.query.ipv6
num.query.ipv6=3
```

4. `unbound-control stats_shm` shows zero IPv6 queries:
```
$ unbound-control stats_shm | grep num.query.ipv6
num.query.ipv6=0
```

I also noted that when both these commands are executed via `watch` command, most of the time `stats_shm` shows incorrect value, but blinks with correct value from time to time:

`$watch -n 0.5 "unbound-control stats_shm | grep ipv6 && unbound-control stats_noreset | grep ipv6"`

Other statistic metrics shows the same behavior, for example:
```
$ unbound-control stats_noreset | grep AD
num.query.flags.AD=3
$ unbound-control stats_shm | grep AD
num.query.flags.AD=0
```

Accessing the shared memory regions with `shmat`/`shmget` results in the same behavior.

I guess it happens because [`shm_main_run`](https://github.com/NLnetLabs/unbound/blob/ff92edcd410610e0f506e4daf0ffbad58ed01760/util/shm_side/shm_main.c#L218) function does not do the memory update atomically and read operations happen during the update process?

wcawijngaards:
Fix statistics with shared memory. The total was added by the threads and thus the partial total is exposed while the threads add up. The thread 0 wipes the total, but is not guaranteed to be the first, so the total contains stats from different time periods. The fix is there to stop that.

The mesh_time_median stat value is fixed to add up to the correct average that is used, as well.

The stat interval is selected so that threads have stat timer callbacks at the same time. The threads use the same time offset in the interval for that.

The stat totals are collected together in a separate struct. When it is done, the struct with totals is copied into the shared memory area.

Fixes #152 . Since the totals are added together properly before the copy to the shared memory. If some threads are not responding, the total does not get updated during that time, but gets updated afterwards.

### DIFF
--- a/daemon/daemon.c
+++ b/daemon/daemon.c
@@ -837,6 +837,12 @@ daemon_fork(struct daemon* daemon)
 	 * the thread_start() procedure.
 	 */
 	set_log_thread_id(daemon->workers[0], daemon->cfg);
+	/* If shm stats need an offset, calculate it */
+	if(daemon->cfg->shm_enable && daemon->cfg->stat_interval > 0) {
+		daemon->stat_time_specific = 1;
+		daemon->stat_time_offset =
+			((int)time(NULL))%daemon->cfg->stat_interval;
+	}
 
 #if defined(HAVE_EV_LOOP) || defined(HAVE_EV_DEFAULT_LOOP)
 	/* in libev the first inited base gets signals */
--- a/daemon/daemon.h
+++ b/daemon/daemon.h
@@ -143,7 +143,14 @@ struct daemon {
 	/** the dnstap environment master value, copied and changed by threads*/
 	struct dt_env* dtenv;
 #endif
+	/** The SHM info for shared memory stats. */
 	struct shm_main_info* shm_info;
+	/** * If the timeout for statistics is attempted at specific offset.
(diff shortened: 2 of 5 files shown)


gthess (approved):
Nice code!
Some typos and text suggestions in https://github.com/NLnetLabs/unbound/pull/1427.

PR #1409, opened 2026-02-21
Documentation - CNAME in redirect-type local-zone

hdais:
CNAME resolution in `redirect` local-zone (`doc/CNAME-basedRedirectionDesignNotes.pdf`) is useful feature but no description in man pages. Please include its description.


wcawijngaards (approved):
Thank you for the additional explanation. The redirect feature works to have CNAMEs from local data have upstream dns resolution.

wcawijngaards:
I added slightly more text to the section, in the commit. About that it has to be at zone apex, and wildcard processing feature details.

PR #1411, opened 2026-02-28
Allow synthesized DNAME TTL=0 to be served from cache within grace period

arunabha007:
## Summary

Addresses the doc/TODO item: "understand synthesized DNAMEs, so those TTL=0 packets are cached properly."

When an authoritative server returns a DNAME with TTL=0 (RFC 2308: do not cache), Unbound previously never reused it for synthesis. Every query under that DNAME caused full recursion.

## Changes

- **services/cache/rrset.c**: Allow TTL=0 DNAME rrsets to be returned from `rrset_cache_lookup` within a 1-second grace period, when they have the `PACKED_RRSET_UPSTREAM_0TTL` flag.
- **services/cache/dns.c**: Update `synth_dname_msg` to accept TTL=0 DNAME rrsets (via `PACKED_RRSET_UPSTREAM_0TTL`) and return TTL=0 in the synthesized response to the client.
- **doc/Changelog**: Add entry for this change.

## Result

- Queries under TTL=0 DNAME can be answered from cache within the 1-second grace window, reducing recursion and load.
- Client-facing behavior is unchanged: responses still return TTL=0 (RFC compliant).
- After the grace period, Unbound fetches fresh data again, as intended for TTL=0.

## Testing
iter_dname_ttl0.rpl — Replay test for DNAME with TTL=0
Purpose: Checks that Unbound handles DNAME records with TTL=0 (RFC 2308 “do not cache”) when they are signed with RRSIG Original TTL=0 (RFC 4034).
Scenario:
Query 1: Resolve foo.test-dname.example.com A. The authority returns DNAME TTL=0, CNAME TTL=0, and an A record.
Query 2: Resolve foo2.test-dname.example.com A (same family, no TIME_PASSES). The client should receive DNAME and CNAME with TTL=0 and the A record.
Details:
Uses DNSSEC-signed DNAME with TTL=0 and RRSIG Original TTL=0 (via ldns-signzone).
Uses trust anchors for example.com and example.net.
Ensures the client gets TTL=0 in responses and that the resolver behaves correctly for TTL=0 DNAME.
Exercises both direct recursion and, when applicable, synthesis from cache during the grace period.

- `make test` passes.
- `make longtest` passes

### DIFF
--- a/doc/Changelog
+++ b/doc/Changelog
@@ -1,6 +1,8 @@
-27 February 2026: Wouter
-	- Merge #1409: Documentation CNAME in redirect-type local-zone.
-	- Update generated man pages.
+27 February 2026: 
+	- Allow synthesized DNAME responses with TTL=0 to be served from cache
+	  within a 1-second grace period (doc/TODO). Reduces recursion when
+	  authoritative servers return DNAME with TTL=0 (RFC 2308). Response
+	  still returns TTL=0 to clients. Test with proper TTL=0 RRSIGs TBD.
 
 25 February 2026: Wouter
 	- Fix validator to set unchecked when validation recursion
--- a/services/cache/dns.c
+++ b/services/cache/dns.c
@@ -766,8 +766,15 @@ synth_dname_msg(struct ub_packed_rrset_key* rrset, struct regional* region,
 		rrset->entry.data;
 	uint8_t* newname, *dtarg = NULL;
 	size_t newlen, dtarglen;
-	if(TTL_IS_EXPIRED(d->ttl, now))
-		return NULL;
+	time_t rr_ttl;
+	if(TTL_IS_EXPIRED(d->ttl, now)) {
+		/* Allow TTL=0 DNAME from upstream within grace period */
+		if(!(rrset->rk.flags & PACKED_RRSET_UPSTREAM_0TTL))
+			return NULL;
+		rr_ttl = 0;
+	} else {
+		rr_ttl = d->ttl - now;
+	}
 	/* only allow validated (with DNSSEC) DNAMEs used from cache 
 	 * for insecure DNAMEs, query again. */
 	*sec_status = d->security;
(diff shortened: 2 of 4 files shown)


arunabha007:
@wcawijngaards request you to check . Thanks

wcawijngaards:
The code looks good! And it includes a the unit test. And I think it should work fine. The grace period for the 0TTL DNAME is fine to have, and avoiding recursion time by allowing its use is very nice to have.

So a question I have, is why to need this. Is there a case where this is used for? Otherwise it is a good solution, but I am trying to see what the case is for it. I mean, do you need it in particular, for something, or is there a place where such 0TTLs are often present, perhaps?

It is also a todo item, thank you for noting the item and working for that! It may have been that I wrote that item down for making DNAMEs picked out of the cache individually, as they already are in the code, also when a 0TTL CNAME was present. But this work on that topic can also be incorporated, but I am trying to see if it is for something in particular.

arunabha007:
Hi @wcawijngaards : Thanks for the review and for clarifying the TODO.

Why this is useful ?
Standards and real-world 0 TTL
-----------------------------------
RFC 2308 allows TTL=0 for “do not cache”. We don’t have a single famous zone that serves DNAME with TTL=0, but:
        - TTL=0 is used in practice (e.g. NSEC3PARAM at .es and in BIND).
        - Authoritative servers can use DNAME with TTL=0 for the same “do not cache” policy (e.g. dynamic DNAME, or to force fresh lookups). The change makes Unbound behave sensibly when they do.
 
What happens without the change?
---------------------------------------
For any such zone, every query under the DNAME triggers full recursion. A short burst of queries (e.g. one page load with several subdomains) causes repeated identical upstream queries. That increases load on the resolver and the authoritative and doesn’t add value, since the answer is unchanged within that second.

What the change does ?
---------------------------
We keep “do not cache” in the RFC sense: we don’t store the full message in the msg cache, and we still return TTL=0 to the client. We only allow synthesis from the already-received TTL=0 DNAME rrset for 1 second. So we reduce redundant recursion without changing the semantics the client sees.

Relation to the TODO
------------------------
Even if the TODO was mainly about “DNAMEs picked from cache when 0 TTL CNAME is present”, this addresses the same idea: making DNAME synthesis work sensibly with 0 TTL so the cache can still be used for synthesis in a minimal, RFC‑compliant way. It fits that theme and closes the item in a small, conservative way.

Personal need
---------------
I don’t have a specific deployment that requires it right now; I implemented it as a general improvement for the TODO and for any current or future use of TTL=0 DNAME. If you’re comfortable with that, I’d be happy to see it merged. :) 👍

wcawijngaards:
I like the protocol compliance, while also reducing recursion need. That improves cache efficiency. Thank you for the implementation!

arunabha007:
@wcawijngaards Thanks for taking the time to review and merge this change. I appreciate your feedback on the design. I’d be glad to contribute again.

PR #1415, opened 2026-03-06
Add lock unlock for view in memory error handling

aww-aww:
view->lock would be skipped during an out of memory error bailout.


wcawijngaards (approved):
Yes, that is needed for an out of memory condition, when respip is used with a view. The change looks good.

wcawijngaards:
That is a nice bugfix! Thank you for the updated code.

PR #1422, opened 2026-03-14
Send along ETag with auth zone xfrs over http(s)

willem-ietf125:
When fetching an authoritative zone over http(s), store the ETag if it was send along in the response, and subsequentally send that along in subsequent http(s) requests as a "If-None-Match: <ETag>| header.

This will prevent Unbound from fetching the full root zone every 30 minutes when configured for local root wiht an url, but will instead only fetch it when there is a different version.

### DIFF
--- a/services/authzone.c
+++ b/services/authzone.c
@@ -5100,7 +5100,7 @@ apply_axfr(struct auth_xfer* xfr, struct auth_zone* z,
 /** apply HTTP to zone in memory. z is locked. false on failure(mallocfail) */
 static int
 apply_http(struct auth_xfer* xfr, struct auth_zone* z,
-	struct sldns_buffer* scratch_buffer)
+	struct sldns_buffer* scratch_buffer, char* etag)
 {
 	/* parse data in chunks */
 	/* parse RR's and read into memory. ignore $INCLUDE from the
@@ -5193,6 +5193,10 @@ apply_http(struct auth_xfer* xfr, struct auth_zone* z,
 				sldns_buffer_begin(scratch_buffer));
 			return 0;
 		}
+	}
+	if (etag) {
+		verbose(VERB_OPS, "etag %s stored", etag);
+		strcpy(z->etag, etag);
 	}
 	return 1;
 }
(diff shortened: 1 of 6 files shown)


services/authzone.c
```diff
@@ -5551,6 +5555,8 @@ xfr_transfer_init_fetch(struct auth_xfer* xfr, struct module_env* env)
 #endif
 
 	if(master->http) {
+		struct auth_zone* z = auth_zone_find(
+			env->auth_zones, xfr->name, xfr->namelen, xfr->dclass);
```
wcawijngaards: For the locks it can be better to have something, that picks up the etag when the zone is locked?
```suggestion
		char etag[128], *etag_str = NULL;
		struct auth_zone* z;
		/* Obtain locks and structures. */
		lock_basic_unlock(&xfr->lock);
		if(!xfr_process_reacquire_locks(xfr, env, &z)) {
			/* The zone is gone, ignore xfr transfer. */
			return 1; /* Stop processing this xfr. */
		}
		/* Holding xfr and z locks. */

		if(z && z->etag[0] && strlen(z->etag) < sizeof(etag)) {
			strlcpy(etag, z->etag, sizeof(etag));
			etag_str = etag;
		}
		lock_rw_unlock(&z->lock);
```

services/authzone.c
```diff
@@ -5560,7 +5566,7 @@ xfr_transfer_init_fetch(struct auth_xfer* xfr, struct module_env* env)
 		xfr->task_transfer->cp = outnet_comm_point_for_http(
 			env->outnet, auth_xfer_transfer_http_callback, xfr,
 			&addr, addrlen, -1, master->ssl, master->host,
-			master->file, env->cfg);
+			master->file, env->cfg, (z && z->etag[0] ? z->etag : NULL));
```
wcawijngaards: This is together with the change that adds locks. Uses the etag_str to pass an etag if there is one.
```suggestion
			master->file, env->cfg, etag_str);
```

services/authzone.c
wcawijngaards: It picks up a write lock, but only needs a read lock. But otherwise it stops a locking issue on the `z->etag` member.

PR #1428, opened 2026-03-30
Centralize scratchpad cleanup in daemon/worker.c

pgporada:
Improve readability by centralizing `regional_free_all` calls preceeding `send_reply` and `send_reply_rc` into just `send_reply_rc`. Adds comments so that editors can get a tooltip.


pgporada:
On second thought, I'm going to close this out.

PR #1430, opened 2026-04-02
Fix race condition in proxy_protocol.c

aww-aww:
We're seeing consistent crashes with sanitization enabled. Fix data race in pp_init by using atomics when multiple threads call it concurrently. Simplify to use local write functions in pp_init. Verified with unbound_context_test.


util/proxy_protocol.c
```diff
@@ -112,22 +120,22 @@ pp2_write_to_buf(uint8_t* buf, size_t buflen,
...
+		(*write_uint32)(buf, 0);
 		buf += 4;
 		/* src port */
 		memcpy(buf,
 			&((struct sockaddr_in*)src)->sin_port, 2);
 		buf += 2;
 		/* dst addr */
 		/* dst port */
-		(*pp_data.write_uint16)(buf, 12);
+		(*write_uint16)(buf, 12)
```
aww-aww, line 138: Should this be buf, 0? 12 is the size of the field, but based on the ipv6 dst port below it should be 0, not the field size?

PR #1433, opened 2026-04-09
Add new static zone type block_aaaa to suppress AAAA queries

### LINKED ISSUE #551: [FR] Option to provide only A/AAAA records
As part of moving our network to IPv6 only, we're using DNS64, which works as expected with unbound when querying AAAA records for sites that only have IPv4 setups.

One challenge in taking the next step is that there does not appear to be a way to have unbound provide only AAAA records. This results in IPv6 queries resulting in A records, and continued pressure/traffic on my IPv4 system, rather than transitioning all traffic other than needed IPv4 traffic to IPv6. I'm trying to have unbound ONLY provide AAAA records.

Perhaps I'm missing a respip setting or some other query option? Respip can filter all A queries (or all AAAA queries), but it doesn't get my dig commands to see the great responses waiting for them in AAAA land. Respip also seems heavy handed, filtering after a response comes back, rather than stopping it before going out.

I can see a situation where individuals may want to restrict responses to only be A records, though that is not my circumstance.

Secondary request: Since 'dns64 subnetcache respip validator iterator' works, can we also assume that 'dns64 respip validator iterator' works and remove the error code? All 'potential error' codes throw failure notices in OpenBSD's rc script, even if it would normally start.

#unbound -V 
Version 1.13.2

Configure line: --enable-allsymbols --with-ssl=/usr --with-libevent=/usr --with-libexpat=/usr --without-pythonmodule --with-chroot-dir=/var/unbound --with-pidfile= --with-rootkey-file=/var/unbound/db/root.key --with-conf-file=/var/unbound/etc/unbound.conf --with-username=_unbound --disable-shared --disable-explicit-port-randomisation --without-pthreads
Linked libs: pluggable-libevent 1.4.15-stable (it uses kqueue), LibreSSL 3.4.1
Linked modules: dns64 respip validator iterator

# cat unbound.conf
# $OpenBSD: unbound.conf,v 1.21 2020/10/28 11:35:58 sthen Exp $

server:
    verbosity: 1

    interface: ::0

    port: 53

    do-ip4: no
    do-ip6: yes

    module-config: "dns64 validator iterator"
    dns64-prefix: 64:FF9B::/96
    access-control: ::1 allow
    access-control: X:X:X:X::/64 allow

    root-hints: "/var/unbound/etc/root.hints"

    hide-identity: yes
    hide-version: yes

    harden-glue: yes

    harden-dnssec-stripped: yes

    use-caps-for-id: yes

    cache-min-ttl: 3600
    cache-max-ttl: 86400
    prefetch: yes
    num-threads: 4
    msg-cache-slabs: 8
    rrset-cache-slabs: 8
    infra-cache-slabs: 8
    key-cache-slabs: 8

    so-rcvbuf: 1m

    unwanted-reply-threshold: 10000

    do-not-query-localhost: no

    val-clean-additional: yes

  forward-zone:
   name: "."
   forward-addr: 2620:fe::fe #Quad9 IPv6
   forward-addr:  2620:fe::9 #Quad9 IPv6-2

jisakiel:
Following d5b9a790f lead for block_a - this would allow suppressing AAAA queries instead for sticking to IPV4.

My use-case in mind is being able to filter certain video providers that consider he.net tunnels as VPNs in a simple*r* way, keeping unbound as my resolver, without needing to involve [python plugins](https://gist.github.com/FiloSottile/e2cffde2bae1ea0c14eada229543aebd), or having to enumerate every subdomain in the `local-data` for a `typetransparent` `local-zone` (as in [here](https://gist.github.com/jamesmacwhite/6a642cb6bad00c5cefa91ec3d742e2a6?permalink_comment_id=4712308#gistcomment-4712308), given the sensible decision of local-data not supporting wildcards). 

It also has a certain nice simmetry with the preexisting `block_a` setting I used as a model.

Checked that it builds, links & passes tests locally only - change seemed straightforward, only being somewhat uncertain on whether something on the response needs to change compared to the A version (I assume not with both being effectively NODATA?). 

I believe this should also close https://github.com/NLnetLabs/unbound/issues/551 for good?

*(and yes, in the year of 2026 there are still ISPs which won't support a proper dual stack configuration, only offering a choice between a non-PD cgnat, or an ipv4-only offering. Shame!)*

### DIFF
--- a/doc/example.conf.in
+++ b/doc/example.conf.in
@@ -896,6 +896,8 @@ server:
 	#   that name
 	# o block_a resolves all records normally but returns
 	#   NODATA for A queries and ignores local data for that name
+	# o block_aaaa similarly to block_a, resolves all records normally but
+	#   returns NODATA for AAAA queries and ignores local data for that name
 	# o always_null returns 0.0.0.0 or ::0 for any name in the zone.
 	# o noview breaks out of that view towards global local-zones.
 	#
--- a/doc/unbound.conf.5.in
+++ b/doc/unbound.conf.5.in
@@ -2920,6 +2920,7 @@ The types are
 \fI\%inform_redirect\fP,
 \fI\%always_transparent\fP,
 \fI\%block_a\fP,
+\fI\%block_aaaa\fP,
 \fI\%always_refuse\fP,
 \fI\%always_nxdomain\fP,
 \fI\%always_null\fP,
@@ -3107,6 +3108,15 @@ ignores local data and resolves normally all query types excluding A.
 For A queries it unconditionally returns NODATA.
 Useful in cases when there is a need to explicitly force all apps to
 use IPv6 protocol and avoid any queries to IPv4.
+.UNINDENT
+.INDENT 7.0
+.TP
+.B block_aaaa 
+Like \fI\%transparent\fP or \fI\%block_a\fP, but
+ignores local data and resolves normally all query types excluding AAAA.
+For AAAA queries it unconditionally returns NODATA.
+Useful in cases when there is a need to explicitly force all apps to
+use IPv4 protocol and avoid any queries to IPv6.
 .UNINDENT
 .INDENT 7.0
 .TP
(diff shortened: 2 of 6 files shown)


wcawijngaards (approved):
The feature looks to work. It is useful, even though it would be nice if IPv6 worked out, in certain situations.

PR #1437, opened 2026-04-16
Attempt to compile under openssl4

pemensik:
Not yet properly tested. Some warnings are not yet solved.


smallapp/unbound-anchor.c
```diff
@@ -1674,6 +1674,11 @@ static unsigned long
 get_usage_of_ex(X509* cert)
 {
 	unsigned long val = 0;
+#ifdef HAVE_X509_GET_EXTENDED_KEY_USAGE
+	val = X509_get_extended_key_usage(cert);
+	if (val == UINT32_MAX)
+		return 0;
+#else
 	ASN1_BIT_STRING* s;
 	if((s=X509_get_ext_d2i(cert, NID_key_usage, NULL, NULL))) {
```
botovq, line 1683: Does unbound still support OpenSSL prior to 1.1.0? If not, one could do this instead:

```
	if((s=X509_get_ext_d2i(cert, NID_key_usage, NULL, NULL))) {
		const unsigned char *data = ASN1_STRING_get0_data(s);
		int len = ASN1_STRING_length(s);

		if(len > 0) {
			val = data[0];
			if(len > 1)
				val |= data[1] << 8;
		}
		ASN1_BIT_STRING_free(s);
```

and if yes, one could probe for `ASN1_STRING_get0_data` and if it's absent `#define ASN1_STRING_get0_data(x) ASN1_STRING_data(x)` then the same code path works for all OpenSSL more recent than SSLeay 0.9.0.

pemensik: Nice catch. That would be working enough for all our version. I think for older OpenSSL we support 1.1 and 3.0+ in addition. Anything older is too old even for us in RHEL.

wcawijngaards:
Than you for the good code suggestions! These are incorporated in the fixup commit that closes the issue. I also fixed the deprecated functions calls. The code makes it through the unit tests, and also works with older versions of OpenSSL.

Issue #1258, opened 2025-03-28
QUIC ipv6 queries fail first try, retry succeeds. ipv4 succeeds


latelatelate:
Forgive me if this is the wrong place, but I'm not sure how to troubleshoot this. I'm running unbound 1.22 compiled w/ QUIC on debian 12 (installed on a proxmox lxc). QUIC queries via IPV6 are failing first try but immediately succeed on retry:

```
root@quic:~# kdig -d ebay.com AAAA @[fd92:7276:c24d::12] -p 2853 +quic
;; DEBUG: Querying for owner(ebay.com.), class(1), type(28), server(fd92:7276:c24d::12), port(2853), protocol(UDP)
;; DEBUG: TLS, received certificate hierarchy:
;; DEBUG:  #1, CN=*.xxx.vg
;; DEBUG:      SHA-256 PIN: 
;; DEBUG:  #2, C=US,O=Let's Encrypt,CN=E5
;; DEBUG:      SHA-256 PIN: 
;; DEBUG: TLS, skipping certificate PIN check
;; DEBUG: TLS, skipping certificate verification
;; WARNING: QUIC, peer took too long to respond
;; WARNING: can't receive reply from fd92:7276:c24d::12@2853(UDP)
;; DEBUG: retrying server fd92:7276:c24d::12@2853(UDP)
;; DEBUG: TLS, received certificate hierarchy:
;; DEBUG:  #1, CN=*.xxx.vg
;; DEBUG:      SHA-256 PIN: 
;; DEBUG:  #2, C=US,O=Let's Encrypt,CN=E5
;; DEBUG:      SHA-256 PIN: 
;; DEBUG: TLS, skipping certificate PIN check
;; DEBUG: TLS, skipping certificate verification
;; QUIC session (QUICv1)-(TLS1.3)-(ECDHE-SECP256R1)-(ECDSA-SECP256R1-SHA256)-(AES-256-GCM)
;; ->>HEADER<<- opcode: QUERY; status: NOERROR; id: 0
;; Flags: qr rd ra; QUERY: 1; ANSWER: 0; AUTHORITY: 1; ADDITIONAL: 1

;; EDNS PSEUDOSECTION:
;; Version: 0; flags: ; UDP size: 1232 B; ext-rcode: NOERROR

;; QUESTION SECTION:
;; ebay.com.                    IN      AAAA

;; AUTHORITY SECTION:
[...]

PR #1441, opened 2026-04-22
Fix buffer overrun in doq_repinfo_retrieve_localaddr()

Fothsid:
SAST finding.

`struct in_addr{,6}` and `struct sockaddr_in{,6}` are different structures, with the latter including the former and generally being bigger. In case of the IPv6 path, the `memmove()` call could write past `struct doq_addr_storage`, most likely overwriting the `addrlen` field of `struct doq_pkt_addr`.

Not sure what the implications are of this error, but I would expect for `doq_lookup_repinfo()` to fail to find anything for IPv6 replies. Tried to look for issues related to QUIC & IPv6 and found only #1258, maybe it's somehow related?


wcawijngaards (approved):
This change looks like a good bugfix. I reviewed other use in Unbound, but those seem to be fine.

Fothsid:
(Whoops, sorry for the typo.)

PR #1451, opened 2026-05-18
Feature: Add nftset module for nftables support

Kasoo:
This PR extends the existing iptables/PF ipsets functionality of Unbound to add support for modern nftables sets.

ipsets/nftsets are often used to allow domain-based firewall rules to be constructed, by having Unbound populate newly resolved IP addresses into the sets.

The specific use-case for this is OpenWRT and PBR (Policy Based Routing) which only support nftsets on their recent versions.

Additionally both nftset and ipset have been given support for per-zone sets (configured with set: "zone" "name-v4" "name-v6"). Previously only a single global set name was supported.

Both backends share the majority of their code in ipset.c and use netlink via libmnl on Linux. The ipset path is unchanged on BSD.

New support has been added to both ipset and nftset for checking netlink errors and logging error messages via the kernel-provided NETLINK_EXT_ACK string (previously errors were ignored).

Finally CAP_NET_ADMIN is now preserved across the privilege drop on Linux when the ipset or nftset module is configured. Previously the ipset module required the process to run as root to function.

AI Disclosure: Code written using Claude/Gemini. All code reviewed and tested by me.


ipset/ipset.c
```diff
@@ -138,12 +270,11 @@ static int add_to_ipset(filter_dev dev, const char *setname, const void *ipaddr,
 	}
 	return 0;
 }
-#else
+#elif defined(USE_IPSET)
 static int add_to_ipset(filter_dev dev, const char *setname, const void *ipaddr, int af) {
 	struct nlmsghdr *nlh;
-	struct nfgenmsg *nfg;
 	struct nlattr *nested[2];
-	static char buffer[BUFF_LEN];
```
Kasoo: Note, these buffers have been switched from static to stack-local due to multi-threaded re-entrancy concerns.

We are now having to stack allocate a fairly big (2K) stack buffer, this is because checking netlink errors requires that we be able to receive the relatively large `NLMSG_ERROR` messages which contains both the original request that errored and the error string.

The stack buffer is reused by `netlink_drain_errors`.

ipset/ipset.c
```diff
@@ -466,15 +773,9 @@ void ipset_inform_super(struct module_qstate *ATTR_UNUSED(qstate),
 }
 
 void ipset_clear(struct module_qstate *qstate, int id) {
-	struct cachedb_qstate *iq;
 	if (!qstate) {
 		return;
 	}
-	iq = (struct cachedb_qstate *)qstate->minfo[id];
```
Kasoo, line 476: I believe this code was left-behind copy-pasted from another module (and didn't do anything)

egc112:
I hope that this will be implemented soon as it will make live much easier for all OpenWRT users wanting domain based routing while using unbound

sppmasterspp:
I want to use Unbound with PBR too.

PR #1458, opened 2026-05-31
Add client-wait-timeout server option

nshopik:
Adds optional `client-wait-timeout: <msec>` server directive. When a client waits for recursion longer than the configured time, Unbound answers it SERVFAIL + Extended DNS Error 22 (No Reachable Authority); recursion continues in the background, so the cache still fills and the next identical query is served normally. Default `0` (disabled) — opt-in, no existing behavior changes. EDE 22 is attached only when `ede:` is enabled; the libunbound callback path returns plain SERVFAIL with no EDE.

When both `client-wait-timeout` and `serve-expired-client-timeout` are set, whichever fires first answers; the other becomes a no-op (no serve-expired priority). `unbound-checkconf` warns when it sits below `serve-expired-client-timeout` and rejects negatives.

New stats counter `num.queries_client_wait_timeout`, per-thread and summed, via `unbound-control stats`.

### DIFF
--- a/daemon/remote.c
+++ b/daemon/remote.c
@@ -821,6 +821,8 @@ print_stats(RES* ssl, const char* nm, struct ub_stats_info* s)
 		(unsigned long)s->svr.num_queries_cookie_invalid)) return 0;
 	if(!ssl_printf(ssl, "%s.num.queries_discard_timeout"SQ"%lu\n", nm,
 		(unsigned long)s->svr.num_queries_discard_timeout)) return 0;
+	if(!ssl_printf(ssl, "%s.num.queries_client_wait_timeout"SQ"%lu\n", nm,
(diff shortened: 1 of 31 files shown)


nshopik:
We run with `infra-cache-max-rtt: 12000` and with 2 retries instead of 5 to avoid NX-bomb where most domains timeout and exhaust outgoing port list. And need additional knob similar what other public resolvers do (google, cloudflare) when they receive such queries and can't resolve it within 4 seconds. Clients do not listen to answers that arrive 10+ seconds latter most of time. 

I did manual testing with 100K qps to see if this behave exactly as expected and don't leak memory. Our intended use is 
`client-wait-timeout: 4000` 
`infra-cache-max-rtt: 12000`

PR #1465, opened 2026-06-25
Add libunbound/remote.h

dag-erling:
Add a shared header containing prototypes for functions that both ends of a remote control connection need to implement.


wcawijngaards:
The change looks okay. Thank you for making the code better! The change hopefully helps in making the build of the code easier.

wcawijngaards (approved):
The code looks okay. The shared header helps the code build.

PR #1467, opened 2026-06-26
daemon: fix DEREF_AFTER_NULL.EX.COND on worker_init

petrvaganoff:

After having been compared to a NULL value at worker.c:2216, pointer 'worker->stat_timer' is passed in call to function 'worker_restart_timer' at worker.c:2319,where it is dereferenced at worker.c:2029.

Fix that stat_timer creation failure in worker_init does not continue with a NULL timer that causes a crash later.


wcawijngaards (approved):
The cleanup is similar to the other out of memory cleanup actions that are taken there. Thank you for dealing with this error condition, so the server would not crash later on dereferencing the NULL `stat_timer` later.

PR #1468, opened 2026-06-27
Normalize domain names in the configuration

dag-erling:
* In the lexer, when returning a string argument, always allocate one more byte than we need.

* In the parser, whenever we expect a host name, pass it to a function that copies the string in place, removing leading dots, deduplicating intermediate dots, and optionally appending a dot at the end if the caller requested an absolute name.  If the original string was empty or consisted entirely of dots, the result is always a single dot.

This means that common mistakes such as forgetting the final dot in the name of a local zone or including a leading dot in the name of a stub or forwarding zone no longer result in Unbound refusing to start or misbehaving.

This was motivated by seeing Unbound fail to start due to an invalid forwarding zone generated by openresolv from data provided by a misconfigured DHCP server.

### DIFF
--- a/util/configlexer.lex
+++ b/util/configlexer.lex
@@ -627,10 +627,13 @@ iter-scrub-promiscuous{COLON}	{ YDVAR(1, VAR_ITER_SCRUB_PROMISCUOUS) }
         LEXOUT(("QE "));
 	if(--num_args == 0) { BEGIN(INITIAL); }
 	else		    { BEGIN(val); }
-        yytext[yyleng - 1] = '\0';
-	yylval.str = strdup(yytext);
+	/* deliberately allocate one extra byte so there is room to add a
+	 * trailing dot if necessary */
+	yylval.str = malloc(yyleng + 1);
 	if(!yylval.str)
 		yyerror("out of memory");
+	memcpy(yylval.str, yytext, yyleng - 1);
+	yylval.str[yyleng - 1] = yylval.str[yyleng] = '\0';
         return STRING_ARG;
 }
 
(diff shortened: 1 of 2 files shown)


wcawijngaards:
I believe that the best way to have these spelling fixes, is to have them as fix suggestions. They are printed when the parser fails on a domain name. And if it then detects a fail case, it can suggest in the error message the suggested improvement. Also, `unbound-checkconf` is the place where further name, spelling, checks can reside. Also if they are tentative, they could be enabled with an option to it to provide further warnings about names that could be misspelled. Then expert users can not use that option, to silence false positive warnings.

Otherwise, if this is added, the spelling fixup becomes a frustration to expert users that have their config altered from what they intended. So, the error printout makes it obvious what is happening, and also only fires when it is wrong, and suggests the fix, but since it is a suggestion, if the fix is also not the intended result, then the operator can intervene at that point to fix the problem appropriately.

So, moving the spelling suggestion fixes to `unbound-checkconf`, or only after the config parser failed for the config read, and it detects a spelling suggestion is available, then printout the suggested fixup for the name. Otherwise this would become an unwelcome surprise for (expert-) operators, that have their configuration altered from what they want.

So, for the code patch, remove the additional byte that is allocated, that is no longer needed, and wastes memory for some memory constrained users. In addition, if it fails to parse the zone name, in a couple of places in the `util/configparser.y`, then, have a detection routine to see if a spelling suggestion exists, and then have the fixup routine fixup (perhaps a copy with the additional byte space) the string, and then printout with the error message, a suggestion, `did you mean the domain '%s'`, something like that.

dag-erling:
I disagree. The motivation for this change is not typos made by users but a situation in which part of the configuration is automatically generated by a DHCP client and the DHCP server is serving incorrect data (which is _extremely_ common for mass-market broadband equipment) so there isn't even a human in the loop. There is no ambiguity here, and nothing to gain from forcing manual intervention to correct a mistake which we can correct automatically.

wcawijngaards:
So for DHCP generated content, there is a different desire than for operator created content. Perhaps an option that can be enabled for the automated case, or some way to detect the issue? But it is nice to know the correction is then always good to go.

PR #1475, opened 2026-07-19
ipsecmod: fix deref on null in ipsecmod-whitelist after OOM

petrvaganoff:
DEREF_OF_NULL.RET.STAT Return value of a function 'rbtree_create' is dereferenced at ipsecmod-whitelist.c:105 without checking for NULL, but it is usually checked for this function (5/6).

In ipsecmod_whitelist_apply_cfg(), the return value of rbtree_create() is not checked for NULL before being used.

Found by the static analyzer Svace (ISP	RAS).


wcawijngaards (approved):
This change looks okay. It checks for the allocation failure and then cleans up later.

wcawijngaards:
Thank you for the fix in Unbound! The change has been merged into the code repository.

PR #1476, opened 2026-07-21
ipsecmod: fix possible deref on null after reply_find_answer_rrset()

petrvaganoff:
Return value of a function 'reply_find_answer_rrset' is dereferenced at ipsecmod.c:438 without checking for NULL, but it is usually checked for this function (10/12).

In my opinion, it is better to add a null check, as was done elsewhere.


wcawijngaards (approved):
The code looks okay. It is defense in depth, since a check on that happens earlier in the ipsecmod state, but it is fine to have. The error looks appropriately handled.

PR #1479, opened 2026-07-24
Fix pthread detection on Solaris 11.4

psumbera:
GCC 16 no longer defines _REENTRANT for -pthread on Solaris because pthreads have been provided by libc since Solaris 10 and Solaris 11.4 headers no longer use the macro.

Do not require _REENTRANT on Solaris. The existing link test checks the pthread API directly and continues to select the same flags with older GCC versions.

Regenerate configure with Autoconf 2.71.

Tested on Solaris 11.4 with GCC 15.2 and GCC 16.1.

### DIFF
--- a/ax_pthread.m4
+++ b/ax_pthread.m4
@@ -87,7 +87,7 @@
 #   modified version of the Autoconf Macro, you may extend this special
 #   exception to the GPL to apply to your modified version as well.
 
-#serial 31
+#serial 32
 
 AU_ALIAS([ACX_PTHREAD], [AX_PTHREAD])
 AC_DEFUN([AX_PTHREAD], [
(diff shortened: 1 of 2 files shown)


wcawijngaards:
Does that work for older solaris versions? Also the link at the top of the file, the current version matches the latest version that is available. Perhaps the change should really go there, or go there as well. Could it perhaps check a solaris version (>11.4 or so?). Other than that, it looks fine.

psumbera:
Thanks. I have updated the PR to support both older and newer Solaris releases and aligned the change with autoconf-archive PR #341  (https://github.com/autoconf-archive/autoconf-archive/pull/341).

Instead of checking the Solaris version at runtime, it preprocesses `<sys/feature_tests.h>` and checks `_XOPEN_VERSION`. On older Solaris releases without `XPG7`, `_REENTRANT` is still required. On Solaris 11.4 with `XPG7`, only the `_REENTRANT` check is skipped; the normal pthread compile and link test still runs.

Checking the target headers also works when cross-compiling, unlike a uname-based Solaris version check.

The `ax_pthread.m4` change now matches the Autoconf Archive proposal. configure was regenerated with Autoconf 2.71.

I tested this with GCC 15.2 and GCC 16.1 on Solaris 11.4. The Autoconf Archive change has additionally been tested on Solaris 11.3.

wcawijngaards:
Yes that works. Merged it. Thanks for improving the portability of the source code! The similarity with the proposed change in the upstream file is good to have.

PR #1493, opened 2026-08-13
pythonmod: check return values of fseek()

petrvaganoff:
Hello!

Added another check for returned values ​​to ensure consistency across checks.

fseek() may fail and leave the file position unchanged. Check its return value before using the resulting file position or reading the file.

Also move the rewind before the allocation in the Python 3.9+ error path so that no additional cleanup is needed if fseek() fails.

Follow-up to #1478.

### DIFF
--- a/pythonmod/pythonmod.c
+++ b/pythonmod/pythonmod.c
@@ -470,7 +470,12 @@ int pythonmod_init(struct module_env* env, int id)
       /* for python before 3.9 */
       log_err("pythonmod: can't parse Python script %s", pe->fname);
       /* print the error to logs too, run it again */
-      fseek(script_py, 0, SEEK_SET);
+      if(fseek(script_py, 0, SEEK_SET) != 0) {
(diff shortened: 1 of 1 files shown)


pythonmod/pythonmod.c
```diff
@@ -504,12 +514,17 @@ int pythonmod_init(struct module_env* env, int id)
...
+            "%s: %s", pe->fname, strerror(errno));
+         goto fail_close_file;
+      }
+
       fstr = malloc(flen+1);
       if(!fstr) {
 		log_err("malloc failure to print parse error");
 		goto fail_close_file;
       }
-      fseek(script_py, 0, SEEK_SET);
```
socketpair, line 512: logically, `fseek()` + `fread()` should go together.

petrvaganoff: Okay, I made them together.

socketpair (commented):
Now ok

PR #1510, opened 2026-09-21
Move unlink of temp file to before fatal exit

aww-aww:
Ensure temporary file is deleted only on failure. As written this would never be executed.


gthess:
Thanks for catching that! It seems to be a left-over from when `log_err()` was changed to `fatal_exit()` there.

